PR #94, opened 2015-10-19
Fix for callback error.

benpatterson:
I did this while troubleshooting an issue (I commented here: https://github.com/sitespeedio/sitespeed.io/issues/745.)  (The issue ended up being browsermob proxy getting hung and not responding. That was due to the paltry resources I had used on an EC2 node.) 

With this change, I was able to get the underlying exception and continue my troubleshooting at the time.

Interested in feedback. Thanks for the look.


benpatterson:
Closing this since it's (a) old, and (b) sitespeed's on 4.x now.

PR #96, opened 2015-11-25
Add user timing measures to stats

msecret:
This adds user timing measures to the stats, using the duration
property as the primary metric number (which differs from the marks.

The measure ability is useful over mark because you can do certain
calculations on multiple marks before creating a measure as well as
being able to measure how long specific parts of the process take.
I can't think of a reason why measures wouldn't be used.


soulgalore:
Fixed in #103

PR #100, opened 2016-01-24
Response Header and Cookie exists check

beenanner:
Wrapping response cookie entries in check to make sure they exist before parsing. This revealed another issue with multiple runs of a decent size site as JSON.stringify on the generating of the har file fails due to the object being too large to work out of memory. I can look into that separately.

### DIFF
--- a/lib/support/chrome-perflog-parser.js
+++ b/lib/support/chrome-perflog-parser.js
@@ -312,9 +312,10 @@ function populateEntryFromResponse(entry, response, timestamp) {
     } else {
       entry.request.headersSize = calculateRequestHeaderSize(entry.request);
     }
-
-    const cookieHeader = response.requestHeaders.Cookie || '';
-    entry.request.cookies = parseCookies(cookieHeader.split(';'));
+    if(response.hasOwnProperty('requestHeaders') && response.requestHeaders.hasOwnProperty('Cookie')) {
+      const cookieHeader = response.requestHeaders.Cookie || '';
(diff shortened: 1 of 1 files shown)


lib/support/chrome-perflog-parser.js
```diff
@@ -312,9 +312,10 @@ function populateEntryFromResponse(entry, response, timestamp) {
     } else {
       entry.request.headersSize = calculateRequestHeaderSize(entry.request);
     }
-
-    const cookieHeader = response.requestHeaders.Cookie || '';
-    entry.request.cookies = parseCookies(cookieHeader.split(';'));
+    if(response.hasOwnProperty('requestHeaders') && response.requestHeaders.hasOwnProperty('Cookie')) {
```
tobli: Only the check for requestHeaders is really needed here. If Cookie is missing, cookieHeader will be set to empty string on the next line.

tobli:
Thanks for fixing the crasher! One stylistic nitpick: I've avoided hasOwnProperty, because AFAIK (I'm certainly no Javascript expert) it doesn't add anything for checking the presence of a property in cases like this. I see now that there are other implications of response.requestHeaders being optional that I've missed, so I'll go on and make further changes to this area in followup commits.

Thanks again, and you're more than welcome to look into memory usage. Feel free to open a separate issue on that, and provide some examples.

beenanner:
Sure no worries thanks for the heads up, so for clarity you would just go with the following for checks?
`
if(response.requestHeaders) {
  ...
}
`

PR #104, opened 2016-01-31
Surfacing connection id for chrome - useful for debugging

beenanner:
Useful for debugging parallel/delayed requests and connections that are reused. Unfortunately it appears the current Firefox(version 44) uses the port for the connection id which isn't as useful.


tobli:
Cool, thanks!

PR #106, opened 2016-02-07
Avoiding null values for onContentLoad and onLoad

beenanner:
On some recent tests I was getting null values for these 2 entries and as it seems to break har viewers

### DIFF
--- a/lib/support/chrome-perflog-parser.js
+++ b/lib/support/chrome-perflog-parser.js
@@ -202,7 +202,9 @@ module.exports = {
           }
 
           let page = pages[pages.length - 1];
-          page.pageTimings.onLoad = Number((data.timestamp - page._timestamp) * 1000)
+          if(data.timestamp && page._timestamp) {
+            page.pageTimings.onLoad = Number((data.timestamp - page._timestamp) * 1000)
+          }
         }
           break;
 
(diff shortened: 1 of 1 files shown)


tobli:
Thanks @beenanner!
It's great to avoid null values, but I think the real bug is some missing book keeping/incorrect assumptions. The current logic assumes the following flow:
1. Page.frameStartedLoading <- page object created
2. Network.requestWillBeSent <- page timestamp set (if first request on page)
3. Page.loadEventFired <- set onLoad on page

Since timestamps on Network.requestWillBeSent and Page.loadEventFired are mandatory, there shouldn't be any null values when comparing them. So just avoiding the null could mask an underlying problem. My suggestion is to do the null checks, but log a warning message if any of the timestamps are null. Then we can merge this PR and avoid generating har files that cant be viewed while continue to debug.

The current algorithm is generating new page for iframes, when it shouldn't. Creating new pages, and then receiving a Page.loadEventFired event before the first request for the new frame could lead to null values. Maybe something else is going on, but we need more logging for these cases (e.g. of frameIds) to know more. There's a FIXME on Page.frameAttached where I think we need to keep track of iframe loading to avoid unnecessary page creation.

beenanner:
@tobli I'm still wrapping my head around the code base, so thank you for the detailed feedback! I think that is a good solution on logging when null is encountered I'll add that right away. I'm not 100% sure why the values were null at the time and I'm glad you confirmed my suspension it was irregular behavior. I was testing with a fairly heavy website http://www.cnet.com/, so I'll try to reproduce with a much lighter site like https://sitespeed.io/ and if I discover anything I'll open a new issue to track that work.

tobli:
Cool! Thanks for diving in to the code. Both code level and higher level documentation is lacking right now. On top of that we're still changing stuff.

I'll have a look at cnet.com when I get the chance. My hunch is that iframes (i.e. ads…) are what's tripping us up, but don't take my word for it.

tobli:
:+1:

PR #107, opened 2016-02-10
Adding results folder and updating .gitignore to exclude any results

beenanner:


tobli:
Cool! Right now this will crash if the results folder isn't present, so make sure to create that first. I'd also prefer to name the folder browsertime-results, since normally it won't be run in the same folder as the source code, so some context would help. This would also align better with what Sitespeed does. If you could update the path for the Chrome perflog dump files that would be great too.
Thanks!

beenanner:
Ugh I told myself to not forget the results folder in my commit and I did. :-p

tobli:
Thanks!

PR #108, opened 2016-02-14
Attempt to fix chrome experimental waterfalls

beenanner:
Taking in account previous connection to calculate start time of any new requests that reuse the connection and also adding options to disable cache

Before and after:
before-vs-after

### DIFF
--- a/lib/support/chrome-perflog-parser.js
+++ b/lib/support/chrome-perflog-parser.js
@@ -84,7 +84,7 @@ module.exports = {
           let entry = {
             cache: {},
             startedDateTime: moment.unix(data.wallTime).toISOString(), //epoch float64, eg 1440589909.59248
-            _startedTimestamp: data.timestamp,
+            _requestWillBeSentTime: data.timestamp,
             _wallTime: data.wallTime,
             _requestId: data.requestId,
             _frameId: data.frameId,
@@ -128,8 +128,8 @@ module.exports = {
             log.warn('Recieved network response for requestId ' + data.requestId + ' with no matching request.');
             continue;
           }
-
-          entry._fullyLoadedTime = (data.timestamp - entry._startedTimestamp) * 1000;
+          entry._responseReceivedTime = data.timestamp;
+          entry._totalRequestTime = (data.timestamp - entry._requestWillBeSentTime) * 1000;
 
           try {
             populateEntryFromResponse(entry, data.response, data.timestamp);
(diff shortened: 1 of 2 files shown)


beenanner:
I'm going to try to refactor this to capture the time and store it as blocking time instead of bumping out the start time. Blocking is where it should be anyways.

tobli:
Great, that makes more sense I think

lib/support/chrome-perflog-parser.js
```diff
@@ -140,6 +141,15 @@ module.exports = {
             log.error('Error parsing response: ' + JSON.stringify(data, null, 2));
             throw e;
           }
+
+          let previousEntry = previousConnectionEntries.get(data.response.connectionId);
+          if(previousEntry) {
+            let startedDateTime = moment(entry.startedDateTime).valueOf();
```
tobli: Avoid converting to and from string representation of timestamp here. Instead use entry._wallTime

lib/support/chrome-perflog-parser.js
```diff
@@ -140,6 +141,15 @@ module.exports = {
...
+            startedDateTime += (previousConnectionEntries.get(data.response.connectionId));
+            entry.startedDateTime = moment.unix(startedDateTime / 1000).toISOString();
```
tobli: Just moment(startedDateTime).toISOString() should work, right?

beenanner: Yep I need to circle back around on this as I was getting some errors and finally got it working with the above lines of code, definitely needs to be cleaned up and refactored. :-p I also discovered a bug with the first of the requests that reuse the initial connection id, so this may not be the final solution. I just wanted to get it in front of you so you knew the direction I was headed for a solution. ;-) I'm still waiting to hear back from the chrome IRC channel to see if anyone has any ideas on if this approach is valid or if there is an easier way to calculate blocking time.

lib/support/chrome-perflog-parser.js
beenanner: :+1:

beenanner:
Getting an issue with redirected URLs, so I'll have to take a look at that tomorrow, but I feel like it's getting closer.

beenanner:
screenshot from 2016-02-17 01 02 43

beenanner:
Rebased these changes into a single commit, so opened #109

parangsjo:
Hello,

Me and my family have visited a very cozy place recently, you may really like it, just take a look at some pics of it <http://thingoquethi.darioscollection.com/xynpck>

Looking forward, pa@rangsjo.com

Issue #102, opened 2016-01-24
HAR collected from Chrome gives wrong start values?


soulgalore:
Run using <code>--experimental.nativeHar</code> and use Chrome.

When we parse the performance log, it seems that most responses starts at the same time (even though we do HTTP/1):
<img width="852" alt="screen shot 2016-01-23 at 19 26 27" src="https://cloud.githubusercontent.com/assets/540757/12534756/5cd0197c-c266-11e5-933d-bbc13d1aa839.png">

But if we check the dev tools generated har (SAVE AS HAR) it looks like this:
<img width="858" alt="screen shot 2016-01-23 at 19 25 58" src="https://cloud.githubusercontent.com/assets/540757/12534760/7a76bb7a-c266-11e5-8480-e87cd1c01471.png">

beenanner:
@soulgalore @tobli it doesn't look like the HAR generated by browsertime has the connection id field. That could help in validating this issue and could be useful for other types of analysis.

https://dvcs.w3.org/hg/webperf/raw-file/tip/specs/HAR/Overview.html#sec-object-types-entries
> connection [string, optional] (new in 1.2) - Unique ID of the parent TCP/IP connection, can be the client port number. Note that a port number doesn't have to be unique identifier in cases where the port is shared for more connections. If the port isn't available for the application, any other unique connection ID can be used instead (e.g. connection index). Leave out this field if the application doesn't support this info.

https://developer.chrome.com/devtools/docs/protocol/1.1/network#response:-object

![image](https://cloud.githubusercontent.com/assets/723985/12700350/5093afac-c7ac-11e5-89bf-69b29ddd5d92.png)

I'll open up a pull-request with that added feature maybe it will give some insight into what's happening with this issue.

beenanner:
Opened #104 to surface the connection ids in chrome.

tobli:
Merged #104 now, but we'll have to iterate on the perflog parsing. Not sure if it's issues with the data, our calculations or our interpretations if their meaning. The Chrome docs aren't super clear to me.

beenanner:
@tobli @soulgalore quick update as I had some time tonight to take a look at this issue. If the connection setup times on a fresh connection(queuing, stalling, dns lookup, etc) are not taken in account that could explain the reason for the request to all appear to start at the same time.

![image](https://cloud.githubusercontent.com/assets/723985/12740418/f1da040c-c940-11e5-919b-e755b33221b1.png)

[...]

PR #109, opened 2016-02-19
Use previous connection id's time as blocking time

beenanner:
@tobli / @soulgalore for #102 It's ugly and needs some clean up, but should work. 

*Note since Chrome has "Stalled" and "Queued"  and and they do not surface either in the logs I had no choice but to report both as "blocking". In Chrome's Dev Tools "stalled" actually pushes the start time out and "queued" actually reports as blocking time for the next request on the shared connection id. IMO  adding to "blocking" time for both makes more sense to me since if they are sharing connection ids they can't do anything until the previous request is complete in HTTP/1. It more clearly shows performance bottlenecks with having too many requests going to a single domain

### DIFF
--- a/lib/support/chrome-perflog-parser.js
+++ b/lib/support/chrome-perflog-parser.js
@@ -438,9 +438,7 @@ function populateEntryFromResponse(entry, previousConnectionIds, entries, respon
 
     entry.time = dns + connect + send + wait + receive;
 
-    console.log(entry.connection);
     let previousConnectionRequestId = previousConnectionIds.get(entry.connection)
-    console.log(previousConnectionRequestId);
     if(response.connectionReused && (entry._frameId !== entry._requestId) && previousConnectionRequestId) {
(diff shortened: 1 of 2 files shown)


beenanner:
Before and after

screenshot from 2016-02-17 01 02 43

tobli:
Great! I'll have a look today. Cleanup can happen post merge.

lib/support/chrome-perflog-parser.js
```diff
@@ -297,6 +302,12 @@ module.exports = {
       }
     }
 
+    entries.sort((a,b) => {
```
tobli: Hmm, what does this do? parseFloat of an entry would just be NaN.

beenanner: Good catch I was using this to sort the entries by the requestIds. They were out of order causing the waterfall to not look right. I'll update to use the requestIds.

tobli:
Thanks!

PR #111, opened 2016-02-24
Fixing empty redirectURL on entries

beenanner:
As well as removing sort as it conflicts with redirects and order seems fine now after several tests


lib/support/chromePerflogParser.js
```diff
@@ -34,6 +34,11 @@ module.exports = {
         headers: parseHeaders(response.headers)
       };
 
+      let locationHeader = entry.response.headers.find((header) => header.name === "Location");
```
tobli: Use getHeaderValue here, since it's case insensitive.

PR #114, opened 2016-02-27
Refactoring HAR generation to be inline with Chromium

beenanner:
Mirroring Chromium HarEntry.js as much as possible. Still needs work on blocking and startDateTime of entries.

[HAREntry.js from Chromium](https://code.google.com/p/chromium/codesearch#chromium/src/third_party/WebKit/Source/devtools/front_end/sdk/HAREntry.js)

### DIFF
--- a/lib/support/chromePerflogParser.js
+++ b/lib/support/chromePerflogParser.js
@@ -70,33 +70,26 @@ module.exports = {
       entry.connection = response.connectionId.toString();
       entry._connectionReused = response.connectionReused;
 
-      if (response.timing) {
-        let blocked = response.timing["dnsStart"];
-        if (blocked < 0.0) {
-          blocked = 0.0
-        }
-        let dns = response.timing["dnsEnd"] - response.timing["dnsStart"];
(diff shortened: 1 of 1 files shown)


tobli:
Thanks. Great to have the Chromium source to compare with.

soulgalore:
One thing: are we sure it's right in chromium? :) I think our hars looked better than Chrome ... lets re-check.

tobli:
True, Chrome isn't the one source of truth. It's good that we can check Firefox as well, take inspiration from both and make it even better. To refine even further I think we need to set up a local test server, with predictable download times, connection throttling etc.

soulgalore:
I thought (but not 100%) that our har shows dns lookup etc that exported har from Chrome misses. But lets check that again. Its cool to compare how chromium do their thing :)

parangsjo:
Hey,

This is just something really astonishing, I'm telling you, you have to take a look <http://thiwofyque.thegoodicecompany.com/xyadzl>

pa@rangsjo.com

PR #115, opened 2016-02-28
Remove entry sorting method for chrome native har builder

beenanner:
Confirmed sorting isn't needed as the waterfalls are displayed using startDateTime which is incorrect. The items should not be sorted by requestId since the numbers aren't sequentially received due to the 6 or so parallel connections with HTTP/1.1 and would be worse with HTTP/2. Redirects can also cause the requestIds to be out of order since the final request can come after newer requestId that didn't have a redirect.

### DIFF
--- a/lib/support/chromePerflogParser.js
+++ b/lib/support/chromePerflogParser.js
@@ -402,20 +402,6 @@ module.exports = {
       }
     }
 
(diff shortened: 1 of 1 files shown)


tobli:
Great, let's get rid of it and work to the startDateTime right.

PR #116, opened 2016-02-28
Fixing blocked time calculations and removing hack for previous calcuations of blocked time

beenanner:
Fixed incorrectly passed argument to firstNonNegative  method and removed previous hack for using unaccounted time as blocking. 

TODO: investigate the remaining unaccounted time for connections reused.

### DIFF
--- a/lib/support/chromePerflogParser.js
+++ b/lib/support/chromePerflogParser.js
@@ -72,7 +72,7 @@ module.exports = {
 
       let timing = response.timing;
       if (timing) {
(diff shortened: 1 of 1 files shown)


tobli:
Thanks! Let's see if we can take some inspiration from Firefox as well (https://github.com/mozilla/gecko-dev/blob/master/devtools/client/netmonitor/har/har-builder.js).
I'll do some post merge cleanup to get rid of the now unused previousConnectionIds map. I prefer removing unused code rather than commenting out.

Issue #118, opened 2016-02-29
Create examples.md documentation for Browsertime 1.0


beenanner:

tobli:
yes please =)

PR #119, opened 2016-03-02
Examples of commands for browsertime

beenanner:
#118 First round to get documentation going along with a few examples.

### DIFF
--- a/docs/examples/README.md
+++ b/docs/examples/README.md
@@ -1,0 +1,8 @@
+Examples
+========
+All examples are under the assumption that the tool was installed via npm. If you have cloned the git repository you will need to substitute "browsertime" with the path to the JS file "bin/browsertime.js" relative to the checked out repository.
+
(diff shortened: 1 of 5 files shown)


tobli:
Looking great! It'd be great if you could tweak some minor things before merge:

1. All examples assumes running from a checked out git repo. If installed via npm (e.g. npm install  -g browsertime@1.0.0-alpha.1) the commands will be run with "browsertime …". Could be good to at least mention initially.
2. Could you have an option value in all cases where it's required, so the examples parse ok. I think that'l be less confusing. I.e. "bin/browsertime.js https://www.sitespeed.io --chrome.mobileEmulation.width -b chrome" would need a width, as well as width and pixelRatio to run).
3. Squashing the commits would be nice.

beenanner:
Will do on all of the above. Keep the feedback coming!

tobli:
Sweet. Thanks Jonathan!

PR #125, opened 2016-04-07
Adding example to basic page for new screenshot option

beenanner:


parangsjo:
Hello,
I was reading an article and suddenly came accross those words, they are just perfect, you can find the article here <http://thaltuzusly.eliotspencer.org/xyqvrw>
All best, pa@rangsjo.com

PR #145, opened 2016-06-01
Renamed pre/postTask to pre/postScripts + load scripts inside the engine

soulgalore:
Renaming pre/postTask to pre/postScript to make it easier to understand what it does. And making sure the CLI and API handles it the same way (fixes https://github.com/sitespeedio/sitespeed.io/pull/966)


lib/core/engine.js
```diff
@@ -21,8 +23,30 @@ const defaults = {
   experimental: {}
 };
 
+function loadPrePostScripts(tasks) {
```
tobli: the argument should be scripts here (and script below).

tobli:
In engine.js we should rename taskData to scriptData, and update the video scripts and test sample scripts accordingly.

tobli:
I'm not 100% sure about loading scripts from disk in engine.js. The initial idea was for all setup like that (e.g. parsing and transforming cli args) to happen outside of the engine to keep the core classes from growing. Forcing the scripts to be a file on disk limits the flexibility of the engine as well. E.g. it can't be used in a web server where you post the data for the engine run. That's why I've eslint config for lib/core explicitly forbids use of the node 'fs' module. Using require to load and parse files sneaks under the radar…

That said, it's a good change over all. Update taskData to scriptData and we'll merge. Then we can revisit the loadPrePostScripts decision before cutting the next alpha release.

soulgalore:
We can add a parsing module in front, we need to change the API to make it cleaner though, so that you don't have to massage the options before you use it? We have that kind of behaviour today for other scripts if I remember correctly. Maybe we can come up with a solution that fixes them both.

soulgalore:
sooo should we merge this first and then change the name of input parameters as we discussed? or should we fix it all in a new PR?

tobli:
Let's get this in first.

parangsjo:
Hi,

I'm working on a new project and need your opinion on some issues, please take a look here <http://thilotrathu.highspeedi.net/xycbcju>

Good wishes, pa@rangsjo.com

PR #146, opened 2016-06-02
Set the result dir to the absolute dir without timestamp and domain

soulgalore:


tobli:
Great!
I noticed this is not in the cli help yet, let's put it there. We should rename it to resultDir (or something similar) instead, since it's actually where the output is stored, not a base dir where folders are created. But that's all for another PR.

PR #147, opened 2016-06-02
Renaming resultBaseDir to resultDir (as it should be)

soulgalore:
And lets have it in the cli.


lib/support/cli.js
```diff
@@ -207,6 +207,9 @@ module.exports.parseCommandLine = function parseCommandLine() {
       default: false,
       describe: 'Use BrowserMobProxy as a proxy to be able to use basicAuth and trafficShapping (but nut H/2)'
     })
+    .option('resultDir', {
+      describe: 'Set result directory for the files produced by sitespeed.io'
```
tobli: sitespeed.io -> browsertime

soulgalore: oops

Issue #134, opened 2016-05-12
Redirects generates multiple HAR pages in Chrome


soulgalore:
At least for http://leovegas.se see more info here:
https://github.com/sitespeedio/sitespeed.io/issues/920

tobli:
This example is a front end (javascript) redirect. What's the preferred way of handling those? Should perhaps all loads associated with a frame be considered as part of the same HAR page?

PR #152, opened 2016-06-13
Fix for js redirect in same frame

beenanner:
Possible fix for #134


soulgalore:
Tested and it works, thanks @beenanner !

PR #159, opened 2016-06-18
Adding device option to connectivity for browsertime

beenanner:
Unrelated to our other work of getting this to work in docker. This allows the user to override the device for the connection throttling if it isn't the default eth0 device.


soulgalore:
thanks for the cleanup @beenanner , I had missed to change to connectivity on a couple of places. I'll try this out tonight.

Issue #180, opened 2016-08-18
Requested size exceeds screen size error [Firefox]


soulgalore:
There's bug in Marionette that if you set the Window size bigger than the screen you get a :"Requested size exceeds screen size" Error. That changes the behavior from Firefoxdriver.

There's a fix but hasn't been released in stable yet:
See https://github.com/mozilla/geckodriver/issues/56

beenanner:
@soulgalore that issue has been resolved in 49 is this patch still needed?

soulgalore:
Aha is that fixed now, I haven't tried it out yet. We should still have a more sensible screen size? We are also waiting on a Selenium patch so it will work with Geckodriver to set the Window position.

soulgalore:
Yes tried it out setting the size larger than the screen and it works!

PR #181, opened 2016-08-23
#180 - increasing screen size to default to 1280x1024

beenanner:
Will update the changelog once we've confirmed this is working as expected.

### DIFF
--- a/lib/support/startVideo.js
+++ b/lib/support/startVideo.js
@@ -7,14 +7,12 @@ const webdriver = require('selenium-webdriver'),
 
 function getFFMpegArgs(mpegPath, options) {
   // calculate the viewport, lets spend some time and make this better
-  let size = '1366x708';
-  if (options.viewPort) {
-    let widthAndHeight = options.viewPort.split('x');
-    size = widthAndHeight[0] + 'x' + (parseInt(widthAndHeight[1]) - 60);
-  }
+  let size = options.viewPort || '1280x1024';
+  let widthAndHeight = size.split('x');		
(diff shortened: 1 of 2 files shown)


lib/support/startVideo.js
```diff
@@ -7,7 +7,7 @@ const webdriver = require('selenium-webdriver'),
 
 function getFFMpegArgs(mpegPath, options) {
   // calculate the viewport, lets spend some time and make this better
-  let size = '1366x708';
+  let size = '1280x1024';
```
tobli: This should be 1280x964, right? Shaving off 60px for browser ui (window chrome). Could be cleaner to do
```
const size = options.viewPort || '1280x1024';
```
and drop the if-statement to shave off the 60px just in one place.

beenanner: oops forgot all about this I can make those changes @tobli :-)

tobli:
@beenanner is this relevant now that #180 is solved? My earlier comment about the if-statement was unclear. I meant removing just the if-statement, not the body. I.e. unconditionally parse remove 60 vertical pixels, either from the viewPort option, or from '1280x1024'.

beenanner:
@tobli I believe this is still relevant. I went ahead and reverted my previous commits that were incorrect and added back 60 pixel removal. Comparing the video produced by master vs this branch it seems better for Firefox. Chrome still seem a bit off centered though. Feel free to test this branch out if you have time.

beenanner:
Don't believe this is relevant anymore :-)

PR #189, opened 2016-09-25
Add chrome-esque summary line to stdout.  Disable with double -q.

moos:
```shell
$ bin/browsertime.js http://sitespeed.io 
[2016-09-25 00:06:41] Running chrome for url: http://sitespeed.io
[2016-09-25 00:06:41] Testing url http://sitespeed.io run 1
[2016-09-25 00:06:47] Testing url http://sitespeed.io run 2
[2016-09-25 00:06:52] Testing url http://sitespeed.io run 3
[2016-09-25 00:06:57] Wrote data to browsertime-results/sitespeed.io/2016-09-25T000640-0700
12 requests, DOMContentLoaded: 0.82s (±0.15), Load: 1.16s (±0.16), rumSpeedIndex: 960 (±109) (3 runs)
```
where ± are the std-dev of the mean.  With `-q`:
```shell
$ bin/browsertime.js http://sitespeed.io  -q
12 requests, DOMContentLoaded: 0.76s (±0.16), Load: 1.06s (±0.18), rumSpeedIndex: 828 (±164) (3 runs)
```
To silence completely, use `-qq`.  Chrome's summary line from network tab:

screen shot 2016-09-25 at 12 08 25 am

### DIFF
--- a/bin/browsertime.js
+++ b/bin/browsertime.js
@@ -27,6 +27,11 @@ function parseUserScripts(scripts) {
     {});
 }
 
+// 1200 -> 1.2
+function fmt(msec, d=2) {
+  return (msec/1000).toFixed(d).replace(/.0+$/,'');
+}
+
 function run(url, options) {
   let dir = 'browsertime-results';
   if (!fs.existsSync(dir)) {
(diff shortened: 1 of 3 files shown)


soulgalore:
Only briefly looked through code, could be that I've missed but you can run browsertime with whatever javascript code you want (so we don't pickup navigation timings etc), so I think we need a check for every timing we wanna use.

bin/browsertime.js
```diff
@@ -27,6 +27,11 @@ function parseUserScripts(scripts) {
     {});
 }
 
+// 1200 -> 1.2
+function fmt(msec, d=2) {
```
tobli: Default arguments doesn't work, since we still support node 4.3. Might want to change that soon, but let's do that separately.

soulgalore:
I'll merge this tonight and do some small fixes @moos so you don't need to do anything, thanks again, this is  great idea 👍

soulgalore:
thanks @moos !!!

moos:
Glad you found it useful!

PR #191, opened 2016-09-29
Expose cli module and added getOptions() method to get cli options

moos:
Allows parent components to easily reuse browsertime yargs cli options:

``` js
var browsertimeCli = require('browsertime').cli;
let browsertimeOptions = browsertimeCli.getOptions({
  group: 'Browser',
  prefix: 'browsertime.',
  forceGroup: true
});

yargs().options(browsertimeOptions ).option(...)
```

Parent component's help:

``` shell
Browser
  --browsertime.timeouts.browserStart                     Timeout when waiting for browser to start, in milliseconds
                                                                                                                 [number] [default: 60000]

  --browsertime.timeouts.pageLoad                         Timeout when waiting for url to load, in milliseconds [number] [default: 300000]
...
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,8 +4,9 @@ const Engine = require('./lib/core/engine'),
   SeleniumRunner = require('./lib/core/seleniumRunner'),
   errors = require('./lib/support/errors'),
   browserScripts = require('./lib/support/browserScript'),
-  logging = require('./lib/support/logging');
+  logging = require('./lib/support/logging'),
+  cli = require('./lib/support/cli');
 
 module.exports = {
-  Engine, SeleniumRunner, errors, logging, browserScripts
+  Engine, SeleniumRunner, errors, logging, browserScripts, cli
 };
--- a/lib/support/cli.js
+++ b/lib/support/cli.js
@@ -48,194 +48,212 @@ function validateInput(argv) {
   return true;
 }
 
-module.exports.parseCommandLine = function parseCommandLine() {
+function getOptions(opt) {
+  var options = {
+    'timeouts.browserStart': {
+      default: 60000,
+      type: 'number',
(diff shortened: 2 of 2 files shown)


tobli:
Thanks for doing this! I've been thinking about something along these lines. I'd propose to do the filtering of "cli-related" (only relevant when browsertime is run standalone) in browsertime, and not in sitespeed. E.g. separating into getOptions and getCliOptions, and only exporting the first of the two. I think that's a cleaner separation of concerns.

tobli:
I'd prefer aliases to not be exported as well. They are also more cli-centric, and might not be a good fit for sitespeed. The approach in https://github.com/sitespeedio/sitespeed.io/pull/1225 to list aliases explicitly is better I think.

moos:
So cli options would be things like 'silent', 'har', 'output', etc?   I agree -- that would ease the container from having to do the filtering.  I can start with the remove list.  What else would you add there?
```js
removeOptions = ['output', 'silent', 'har', 'skipHar', 'resultDir', 'config'];
```
As for aliases, there is the benefit of continuing/consistency in keeping the same aliases (e.g. `-n-`, `-b`).  But it could lead to conflicts.  Perhaps a `keepAliases` option?

tobli:
Great list, I'd add prettyPrint as well. Let's keep aliases as they are now in the PR, and we can revisit it later if there's a potential conflict. Thanks!

soulgalore:
@tobli is there anything more that needs to be done here or can we merge it? Would be great to get this in and then see if we can reuse the pattern for new (standalone) plugins.

soulgalore:
hey @moos 
do you have time to fix the conflict? Working on fixing as much as possible before 4.0 next week :)

Best
Peter

soulgalore:
I don't think this has a high prio, let us close this and if we ever feel we need re-open.

PR #196, opened 2016-10-13
Increase the resource timing buffer, lets check what's a valid maxSize

soulgalore:


soulgalore:
I'll think I'll remove this PR and then automatically add it the stack when we add a video?

soulgalore:
@beenanner what about setting it to 500 and then have a warning in the logs if we reach that limit?

beenanner:
Yep 500 sounds like a good start. I have tests running so I can monitor the effects on some of the stats ("fully loaded") when bumping that up as well as the getting of the wonderful video on the next release. :-)

PR #197, opened 2016-10-17
Log the total size of the page using the HAR

soulgalore:
Also broke out the creation of the logline to a utility method.

### DIFF
--- a/lib/core/engine.js
+++ b/lib/core/engine.js
@@ -286,8 +286,11 @@ class Engine {
           result.statistics = statistics.summarizeDeep(options);
         }
       }).tap((result) => {
-        log.info(util.getResultLogLine(result,options));
-        return result;
+          const logText = util.getResultLogLine(result);
+          if (logText !== '') {
+            log.info(logText);
+          }
+          return result;
       })
       .catch(Promise.TimeoutError, (e) => {
         throw new UrlLoadError('Failed to load ' + url + ', cause: ' + e.message, url, {
--- a/lib/core/engine.js
+++ b/lib/core/engine.js
@@ -286,25 +286,7 @@ class Engine {
           result.statistics = statistics.summarizeDeep(options);
         }
       }).tap((result) => {
-        // don't bother if no statistics or silent x2
-        if (!result.statistics || !result.statistics.timings || !result.statistics.timings.pageTimings || options.silent > 1) return result;
-
-        let run = result.browserScripts[0].timings,
-          nRuns = result.browserScripts.length,
-          pt = result.statistics.timings.pageTimings,
-          t = result.statistics.timings,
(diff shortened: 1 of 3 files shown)


lib/core/engine.js
```diff
@@ -286,25 +286,7 @@ class Engine {
...
-            util.formatMetric('firstPaint', t.firstPaint, m),
-            util.formatMetric('DOMContentLoaded', pt.domContentLoadedTime, m),
-            util.formatMetric('Load', pt.pageLoadTime, m),
-            util.formatMetric('rumSpeedIndex', t.rumSpeedIndex, m),
-          ],
-          note = m ? ` (${nRuns} runs)` : '';
-
-        lines = lines.filter(Boolean).join(', ');
-        log.info(`${lines}${note}`);
+        log.info(util.getResultLogLine(result,options));
```
tobli: I think the options.silent check belongs here, it's the callers responsibility. Additionally a check for an empty result, to avoid logging undefined/an empty row.

lib/support/util.js
```diff
@@ -14,5 +16,45 @@ module.exports = {
           formatted += ` (±${fmt(metric.mdev)})`;
       }
       return formatted;
+  },
+  getResultLogLine(result, options) {
+    // don't bother if no statistics or silent x2
+    if (!result.statistics || !result.statistics.timings || !result.statistics.timings.pageTimings || options.silent > 1) return result;
```
tobli: Since we now have data to output even if there's just har data (and no statistics) I don't think this check is relevant. Even if it is, it should return an empty string.

lib/support/util.js
```diff
@@ -14,5 +16,45 @@ module.exports = {
...
+
+    let totalSize = 0;
+    let requests = result.browserScripts[0].timings.resourceTimings.length + 1;
+
+    if (result.har && result.har.log.pages[0] !== undefined) {
```
tobli: Could we drop the log.pages[0] check? If we're including malformed hars in the result we should fix that bug elsewhere, rather than checking for it here.

lib/support/util.js
```diff
@@ -14,5 +16,45 @@ module.exports = {
...
+      // get the id
+      let pageId = result.har.log.pages[0].id;
+      let resourcesByPage = groupBy(result.har.log.entries, 'pageref');
```
tobli: maybe entriesByPage?

lib/core/engine.js
soulgalore: seeing now the the silent is a thing from the past (when we used console.log for output), I'll just remove that and make a check for empty message.

lib/support/util.js
soulgalore: Yes!!!!

PR #200, opened 2016-10-30
Replaced continue inside switch case statements with break

shubhamgoyal:
acorn fails with unsyntactic continue otherwise. This is the relevant discussion related to this - https://github.com/ternjs/acorn/issues/58

In this PR, replacing continue with break statements has no possible regression, because there is no code in the body of the loop following the switch expression.

### DIFF
--- a/lib/support/chromePerflogParser.js
+++ b/lib/support/chromePerflogParser.js
@@ -169,7 +169,7 @@ module.exports = {
 
           if (rootFrameMappings.has(frameId) || previousFrameId) {
             // This is a sub frame, there's already a page for the root frame
-            continue;
+            break;
           }
 
           currentPageId = 'page_' + (pages.length + 1);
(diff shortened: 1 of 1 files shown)


shubhamgoyal:
@beenanner @tobli I was getting the following error before I made this change -

> Module parse failed: /Users/shubhamgoyal/performance-measurement-client/node_modules/babel-loader/index.js?presets[]=es2015!/Users/shubhamgoyal/performance-measurement-client/browsertime/lib/support/chromePerflogParser.js Unsyntactic continue (194:40)Module parse failed: /Users/shubhamgoyal/performance-measurement-client/node_modules/babel-loader/index.js?presets[]=es2015!/Users/shubhamgoyal/performance-measurement-client/browsertime/lib/support/chromePerflogParser.js Unsyntactic continue (194:40)

Then, I googled around and found evidence to support this as you can see in the link in my first comment on the PR. It is better to have something that all parsers support than have some people face these issues.

beenanner:
@shubhamgoyal  I agree. I wasn't trying to suggest this is not an appropriate patch it's just that I wanted to confirm with @tobli the initial implementation details for my own sanity. ;-) Thanks again for the patch.

tobli:
Thanks for the PR! The continue is there to skip the event (go to the next in the for loop). As currently written, a break should have the same effect. However that's just because there's no extra code after the switch in the for loop. Break would only break out of the switch, right? Hence I think just replacing continue with break is too brittle, especially since it's not an actual bug in browsertime unless I'm mistaken.

shubhamgoyal:
@tobli I understood the intent. But it fails for me. If you need evidence, I can push a new project which you can clone and do npm install + npm run on. Then, you can see the error for yourself. Let me know if that will help convince you :)

tobli:
@shubhamgoyal yeah, an example would be very helpful. I did not understand what issue you are seeing. Are you using browsertime as a module? How are you expecting Browsertime to work, versus how is it working?

shubhamgoyal:
@tobli Yes, I am using it as a module. I will link to an example later.

soulgalore:
Closing this becasue of inactivity but please let us know if we should re-open.

PR #201, opened 2016-10-31
Cli options

beenanner:
This fixes the conflicts for #191


beenanner:
@tobli  / @soulgalore  anything else needed before merging these changes?

soulgalore:
I think I've missed adding a comment before. I think we should find a way to export the validateInput too (or do we do that already?)?

beenanner:
Good question let me check on that @soulgalore.

beenanner:
@soulgalore you are correct validateInput doesn't work in the sitespeedio branch with this change so we need to find a good way to export those as well.

soulgalore:
I'm not sure we should do this. I think it is ok like it is today and I think it would be better to focus on a generic solution for plugins in sitespeed.io. I wanna close this one and the related https://github.com/sitespeedio/browsertime/pull/191 ok @tobli @beenanner ?

soulgalore:
I don't think this has a high prio, let us close this and if we ever feel we need re-open.

PR #208, opened 2016-11-05
Adding initator to chrome har for use of associating a request with it's caller

beenanner:
@tobli  / @soulgalore this will surface a small object in the chrom HAR on each entry that shows where the request originated from. Possible could be used for something like https://github.com/sitespeedio/pagexray/issues/22

```
...
                _frameId": "27207.1",
                "initiator": {
                    "lineNumber": 0,
                    "type": "parser",
                    "url": "https://www.sitespeed.io/"
                },
                "pageref": "page_1",
                "request": {
                    "method": "GET",
                    "url": "https://www.sitespeed.io/img/sitespeed-logo-2c.png",
...
```

Not 100% sure if this can be done via the Firefox HAR exporter but I know it's available in their devtools in the browser.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Browsertime changelog
 UNRELEASED
 -------------------------
+### Added
+* Added initiator of each request entry to chrome HAR
+
 ### Fixed 
 * Entry timings in HAR files from Chrome were strings instead of numbers.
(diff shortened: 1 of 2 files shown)


beenanner:
Good catch @soulgalore!

beenanner:
Doesn't appear this is available yet in Firefox so adding to chrome only.

Issue #209, opened 2016-11-05
TypeError: Cannot read property '_requestId' of undefined from chromePerflogParser.js


dylangrafmyre:
I am unable to attempt Chrome Version 54.0.2840.87, 3 runs without crashing with ` TypeError: Cannot read property '_requestId' of undefined`. I consistently crash on the first run. 

Just testing against `www.friendsandguests.com` without the custom login works fine, but using the custom login and watching the specific url render properly in the browser seems all is well until it fails within the `chromePerflogParser` module.

Any ideas what I am doing wrong? 

Thanks for the great set of tools!

```
➜  dylan@DYLANs-MacBook-Pro-3:~/sitespeed  sitespeed.io --preScript login.js -vv -b chrome  https://www.friendsandguests.com/s                                     1 ↵ [20:33:24]
[2016-11-04 20:37:02] Versions OS: darwin 14.5.0 nodejs: v6.2.2 sitespeed.io: 4.0.3 browsertime: 1.0.0-beta.9 coach: 0.28.4
[2016-11-04 20:37:02] Starting chrome for analysing https://www.friendsandguests.com/s 3 time(s)
[2016-11-04 20:37:02] Testing url https://www.friendsandguests.com/s run 1
[2016-11-04 20:37:11] Requesting url https://www.friendsandguests.com/s
[2016-11-04 20:37:14] Waiting for script 'return (function() {try { var end = window.performance.timing.loadEventEnd;return (end > 0) && (Date.now() > end + 2000);} catch(e) {return true;}})()' at most 300000 ms
[2016-11-04 20:37:16] Executing script userAgent
[2016-11-04 20:37:16] Executing script windowSize
[2016-11-04 20:37:16] Executing script documentHeight
[2016-11-04 20:37:16] Executing script documentTitle
[2016-11-04 20:37:16] Executing script documentWidth
[2016-11-04 20:37:16] Executing script responsive
[2016-11-04 20:37:16] Executing script firstPaint
[2016-11-04 20:37:16] Executing script fullyLoaded
[2016-11-04 20:37:16] Executing script navigationTiming
[2016-11-04 20:37:16] Executing script pageTimings
[2016-11-04 20:37:16] Executing script resourceTimings
[2016-11-04 20:37:16] Executing script rumSpeedIndex
[2016-11-04 20:37:16] Executing script userTimings
[2016-11-04 20:37:16] Executing script coachAdvice
[2016-11-04 20:37:17] Getting performance logs from Chrome
[...]

PR #211, opened 2016-11-06
Add conditional assignment to handle entries.find returning no value

dylangrafmyre:
If entries.find method call does not find a match, previousEntry will
be undefined causing browsertime to crash.

Added conditional assignment to assign the current entry to
previousEntry if previousEntry === undefined.

#209


tobli:
Thanks for the PR! I think https://github.com/sitespeedio/browsertime/pull/212 is a better fix (I'd love some help to test it out) for reasons stated in #209.

dylangrafmyre:
@tobli Sounds good. I will test and report back. Thanks!

PR #213, opened 2016-11-07
Fixed the key for setting up proxy for HTTPS connections

shubhamgoyal:
I am not sure why the key was set to https. And, since there are no tests for this, I can't verify whether this worked previously. This is quite a simple fix.

### DIFF
--- a/lib/core/webdriver/index.js
+++ b/lib/core/webdriver/index.js
@@ -26,7 +26,7 @@ module.exports.createWebDriver = function(options) {
     let proxyUrl = proxyHost + ':' + options.proxyPort;
     proxyConfig = proxy.manual({
       http: proxyUrl,
-      https: proxyUrl
+      ssl: proxyUrl
     });
   }
 


tobli:
Hi @shubhamgoyal!
The key for configuring a proxy for https is 'https', according to https://github.com/SeleniumHQ/selenium/blob/selenium-3.0.0/javascript/node/selenium-webdriver/lib/proxy.js#L50. So changing to 'ssl' would effectively just disable that setting. Could you describe the problem you were experiencing. The proxy code has admittedly not seen a lot of testing, so a use case would be great!

shubhamgoyal:
@tobli  I set up a proxy for you. I think the best way to demonstrate this is through an example. You can try using this with both https and ssl keys. I am using it right now and it only worked when I changed the key from https to ssl.

Port: 8001
Host: ec2-54-169-99-241.ap-southeast-1.compute.amazonaws.com

soulgalore:
Thinking the best long time solution would be if we could spin up a proxy in Travis so we always automatically test the proxy functionality (both for Chrome/FF)?

tobli:
I tested with a https proxy from https://www.sslproxies.org/, and it worked great. The issue it seems is when using proxies that only support http to the client. That's why changing the https key solves it for @shubhamgoyal, since that just disables the https proxy setting.

So either we add a new property (e.g. --httpsProxy=false), or rework the proxy options with separate options for http and https (and potentially socks as well). I'm closing this PR now, so we can open a new issue for that. Thanks for reporting @shubhamgoyal !

tobli:
@shubhamgoyal e6e4eea38813562551afd2758ba1dd260d6dea3f changed how proxies are configured. Try setting just --proxy.http. I've verified it with the proxy you set up.

PR #214, opened 2016-11-08
Remove outdated Firefox workaround for window position.

tobli:
This doesn’t crash anymore, so remove the hack. Unfortunately it
doesn’t seem to work correctly in Firefox. Setting the window position
doesn’t take effect in my testing.

### DIFF
--- a/lib/core/seleniumRunner.js
+++ b/lib/core/seleniumRunner.js
@@ -63,12 +63,7 @@ class SeleniumRunner {
           viewPort = viewPort.split('x');
 
           const window = this.driver.manage().window();
-          // Do not set position until this is fixed in Geckodriver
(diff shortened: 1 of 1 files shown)


tobli:
ping @beenanner @soulgalore. This should be ok, right?

soulgalore:
ok lets see if I can find the original issue in Geckodriver.

tobli:
@soulgalore it's https://github.com/mozilla/geckodriver/issues/113, and I added a comment to that as well.

soulgalore:
yes but you wrote that it works? :)

tobli:
The command is inplemented in geckodriver. So it doesn't crash anymore, so no need for a conditional workaround. However the window position doesn't seem to be affected in my testing.

Issue #202, opened 2016-11-01
Verify and test SpeedIndex in different viewports


soulgalore:
Check that the video is ok in both Chrome & Firefox and that SpeedIndex and visual start looks ok

soulgalore:
Setting a viewport as an apple device don't give the right speedindex. Check the different timings:
<pre>
docker run --privileged --shm-size=512m --rm -v "$(pwd)":/browsertime sitespeedio/browsertime https://www.sitespeed.io -n 1 --experimental.video true --chrome.mobileEmulation.deviceName 'Apple iPhone 6' -c cable --connectivity.engine tc
Google Chrome 54.0.2840.71 
Mozilla Firefox 49.0.2
Starting Xvfb ...
[2016-11-09 07:52:28] Running chrome for url: https://www.sitespeed.io
[2016-11-09 07:52:28] sudo tc qdisc add dev eth0 root netem delay 28ms loss 0% rate 5000kbps
[2016-11-09 07:52:28] Testing url https://www.sitespeed.io run 1
[2016-11-09 07:52:35] 9 requests, 98.49 kb, firstPaint: 0.27s, firstVisualChange: 2s, DOMContentLoaded: 0.24s, Load: 0.s, speedIndex: 2148, rumSpeedIndex: 295
[2016-11-09 07:52:35] Wrote data to browsertime-results/www.sitespeed.io/2016-11-09T075228+0000
[2016-11-09 07:52:35] sudo tc qdisc del dev eth0 root

</pre>

soulgalore:
We should think about moving xvfb inside the app instead, then we can set the right size of the screen. Or is it a better way to do it?

<img width="1061" alt="screen shot 2016-11-09 at 09 34 05" src="https://cloud.githubusercontent.com/assets/540757/20131890/9724de4e-a660-11e6-8a1d-514b3c987e9a.png">

soulgalore:
This is fixed now.

PR #219, opened 2016-11-09
Xvfb in node

soulgalore:
Use the viewPort to set the window size for XVFB, makes SpeedIndex work for different viewports.

We should make sure that our Docker containers always runs --xvfb. #202

### DIFF
--- a/bin/browsertime.js
+++ b/bin/browsertime.js
@@ -13,8 +13,7 @@ let Engine = require('../').Engine,
   pick = require('lodash.pick'),
   fs = require('fs'),
   path = require('path'),
-  log = require('intel'),
-  Xvfb = require('xvfb');
+  log = require('intel');
 
 Promise.promisifyAll(fs);
 
@@ -49,22 +48,6 @@ function run(url, options) {
     scriptsByCategory = Promise.join(scriptsByCategory, userScripts,
       (scriptsByCategory, userScripts) => merge(scriptsByCategory, userScripts));
   }
-
-let xvfb;
-if (options.xvfb) {
-    const width = options.chrome ? options.chrome.mobileEmulation.width : undefined;
-    const height = options.chrome ? options.chrome.mobileEmulation.height: undefined;
-
-    let widthAndHeight = (width && height) ? width + 'x' + height : options.viewPort;
-
-    xvfb = new Xvfb({
-        displayNum: '99',
-        silent: false,
-        reuse: true,
-        xvfb_args: ['-ac', '-nolisten', 'tcp', '-screen', 0, widthAndHeight + 'x16']
-    });
-    xvfb.startSync();
-}
 
 engine.start()
     .then(function() {
@@ -113,12 +96,7 @@ engine.start()
     .catch(function() {
       process.exitCode = 1;
     })
-    .finally(() => {
-      if (xvfb) {
-        xvfb.stopSync();
-      }
-      process.exit;
-    }); // explicitly exit to avoid a hanging process
+    .finally(process.exit); // explicitly exit to avoid a hanging process
 }
 
 let cliResult = cli.parseCommandLine();
--- a/bin/browsertime.js
+++ b/bin/browsertime.js
@@ -13,7 +13,8 @@ let Engine = require('../').Engine,
   pick = require('lodash.pick'),
   fs = require('fs'),
   path = require('path'),
-  log = require('intel');
+  log = require('intel'),
+  Xvfb = require('xvfb');
 
 Promise.promisifyAll(fs);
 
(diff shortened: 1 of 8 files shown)


bin/browsertime.js
```diff
@@ -49,7 +50,23 @@ function run(url, options) {
...
-  engine.start()
+let xvfb;
+if (options.xvfb) {
+    const width = options.chrome ? options.chrome.mobileEmulation.width : undefined;
+    const height = options.chrome ? options.chrome.mobileEmulation.height: undefined;
+
+    let widthAndHeight = (width && height) ? width + 'x' + height : options.viewPort;
+
+    xvfb = new Xvfb({
+        displayNum: '99',
```
tobli: Do we need to make this configurable? For a Docker container we can be sure there's no conflicts, but maybe not when run outside of Docker.

bin/browsertime.js
```diff
@@ -49,7 +50,23 @@ function run(url, options) {
       (scriptsByCategory, userScripts) => merge(scriptsByCategory, userScripts));
   }
 
-  engine.start()
+let xvfb;
+if (options.xvfb) {
+    const width = options.chrome ? options.chrome.mobileEmulation.width : undefined;
+    const height = options.chrome ? options.chrome.mobileEmulation.height: undefined;
```
tobli: This will crash if --chrome.mobileEmulation.* isn't set. You could use lodash.get, and get(options, 'chrome.mobileEmulation.height')

package.json
```diff
@@ -8,6 +8,7 @@
     "execa": "0.4.0",
     "fast-stats": "0.0.3",
     "hasbin": "1.2.3",
+    "headless": "1.0.0",
```
tobli: Is this needed? It doesn't seem to be used.

lib/support/cli.js
```diff
@@ -236,11 +236,17 @@ module.exports.parseCommandLine = function parseCommandLine() {
       config: 'config'
     })
     .option('viewPort', {
-      describe: 'Size of browser window WIDTHxHEIGHT'
+      describe: 'Size of browser window WIDTHxHEIGHT',
+      default: '1500x1200'
```
tobli: This will always set the window size to 1500x1200, even when not running xvfb. Do we want that as a side effect? Maybe better to just apply it when running xvfb.

bin/browsertime.js
soulgalore: I think it means it will start start with 99 is that one is free? lets check ffmpeg config.

lib/support/cli.js
soulgalore: you are right I'm moving this (we have it set in sitespeed.io though).

package.json
soulgalore: removed!

soulgalore:
There's one more fix we should do but not 100% related to this:
<pre>
[2016-11-09 19:39:48] 9 requests, 98.50 kb, firstPaint: 0.21s (±0s), firstVisualChange: 0.26s (±0.01s), DOMContentLoaded: 0.19s (±0s), Load: 0.29s (±0.01s), speedIndex: 266 (±7.542472332656507), rumSpeedIndex: 214 (±1.9293061504650375) (3 runs)</pre>

lib/core/engine.js
```diff
@@ -60,18 +62,50 @@ class Engine {
...
+      return this.xvfb.stopAsync();
+    }
+    else return Promise.resolve();
   }
 
   start() {
     const runDelegate = this.runDelegate;
       if (this.options.connectivity && this.options.connectivity.profile !== 'native' && this.options.connectivity.profile !== undefined) {
-    return connectivity.set(this.options).then(function() {
+    return this.startXvfb().then(connectivity.set(this.options)).then(function() {
```
tobli: then takes a function, not a promise, as argument. So this should be:

return this.startXvfb()
           .then(() => connectivity.set(this.options))
           .then(function() {...

lib/core/engine.js
```diff
@@ -60,18 +62,50 @@ class Engine {
...
       return runDelegate.onStart();
     }).catch(function(e) {
       log.error('Error running browsertime', e);
       throw e;
     });
-  } else return runDelegate.onStart();
+  } else return this.startXvfb().then(runDelegate.onStart());
```
tobli: Same as above for then

lib/core/engine.js
```diff
@@ -298,12 +332,12 @@ class Engine {
   stop() {
     const runDelegate = this.runDelegate;
     if (this.options.connectivity && this.options.connectivity.profile !== 'native' && this.options.connectivity.profile !== undefined) {
-      return connectivity.remove(this.options).then(function() {
+      return this.stopXvfb().then(connectivity.remove(this.options)).then(function() {
```
tobli: Same as above for then

lib/core/engine.js
```diff
@@ -298,12 +332,12 @@ class Engine {
...
         return runDelegate.onStop()
       });
 
     }
-      else return runDelegate.onStop();
+      else return this.stopXvfb().then(runDelegate.onStop());
```
tobli: Same as above for then

Issue #136, opened 2016-05-15
Add CLI options to remove videos and all files needed for VisualMetrics


soulgalore:
It would be good to just add an option to drop the video and images used for creating SpeedIndex.

PR #225, opened 2016-11-20
Option to remove the video #136

soulgalore:


soulgalore:
Yes thanks. I'll remove the else. About tap, I think it's a native promise we get back from execa right and not bluebird? We should probably have stricter guidance how to handle promises, so we always have the same as much as possible.

tobli:
True! Always wrap in Promise.resolve at the top if non-bluebird. That's the way to get bluebird's warnings about "lost" promises and other goodies.

PR #226, opened 2016-11-22
Add preURL that will be accessed before the page that you wanna analyze

soulgalore:


lib/support/preURL.js
```diff
@@ -0,0 +1,13 @@
...
+
+const Promise = require('bluebird'),
+  log = require('intel');
+
+module.exports = {
+  run(context) {
+      return context.runWithDriver((driver) => {
+          log.info('Accessing preURL %s', context.options.preURL);
+          return driver.get(context.options.preURL);
+        }).then(() => Promise.delay(1500));
```
tobli: What's the reason for the delay here?

tobli: aha, right. So basically it's loadAndWait. Hmm, maybe we should expose the SeleniumRunner to pre/post scripts as well, when you don't need the low level access to the raw selenium WebDriver. So content.runWithSeleniumRunner((runner => runner.loadAndWait(context.options.preURL)) instead.

soulgalore: yes that could probably do that! we still need that delay after the orange page though.

PR #258, opened 2017-01-25
Feature adding whitelist option for usertimings

beenanner:
@tobli / @soulgalore maybe I should move this code to the utils.js or filters.js, so we can keep engine.js small? Thoughts?

### DIFF
--- a/lib/support/cli.js
+++ b/lib/support/cli.js
@@ -273,6 +273,9 @@ module.exports.parseCommandLine = function parseCommandLine() {
     .option('preURL', {
       describe: 'A URL that will be accessed first by the browser before the URL that you wanna analyze. Use it to fill the cache.'
     })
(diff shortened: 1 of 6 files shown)


beenanner:
Actually went ahead and moved it to filter.js :-p

lib/support/cli.js
```diff
@@ -273,6 +273,9 @@ module.exports.parseCommandLine = function parseCommandLine() {
     .option('preURL', {
       describe: 'A URL that will be accessed first by the browser before the URL that you wanna analyze. Use it to fill the cache.'
     })
+    .option('userTimingWhitelist', {
+      describe: 'A regex that will whitelist which userTimings to capture in results.'
```
soulgalore: I think we could just add what happens per default if we leave it out so that is clear.

PR #259, opened 2017-02-04
Use Chromes meaningful paint

soulgalore:
I guess it is still under development but both WPT and Lighthouse uses it heavily so it should be ok to expose.

### DIFF
--- a/lib/support/chromePerflogParser.js
+++ b/lib/support/chromePerflogParser.js
@@ -162,10 +162,7 @@ module.exports = {
       entries = [],
       currentPageId,
       blinkUserTiming = {
-        firstImagePaint: 0,
-        firstContentfulPaint: 0,
         firstMeaningfulPaint: 0,
-        firstTextPaint: 0,
         navigationStart: 0};
 
     for (let event of events) {
--- a/lib/support/chromePerflogParser.js
+++ b/lib/support/chromePerflogParser.js
@@ -474,6 +474,15 @@ module.exports = {
           }
           break;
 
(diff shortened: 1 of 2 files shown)


soulgalore:
If we merge this next step would be to make the trace categories configurable to make https://github.com/sitespeedio/sitespeed.io/issues/1423 possible.

lib/support/chromePerflogParser.js
```diff
@@ -474,12 +477,23 @@ module.exports = {
           }
           break;
 
+        case 'Tracing.dataCollected':
+          if (blinkUserTiming[message.params.name] === 0) {
+            blinkUserTiming[message.params.name] = message.params.ts;
```
tobli: This can set on pages[pages.length - 1], just like Page.loadEventFired

lib/support/chromePerflogParser.js
```diff
@@ -474,12 +477,23 @@ module.exports = {
...
+
         default:
           log.warn('Unhandled event: ' + message.method);
           break;
       }
     }
 
+    // lets add firstMeaningFulPaint for now, the other are not ok
+    if (blinkUserTiming.firstMeaningfulPaint - blinkUserTiming.navigationStart > 0) {
+      pages[0].pageTimings._firstMeaningfulPaint = (blinkUserTiming.firstMeaningfulPaint - blinkUserTiming.navigationStart) / 1000;
```
tobli: Loop and set per page.

soulgalore:
Skip this since the meaningful paint is still quite off.

PR #264, opened 2017-02-09
include response bodies in HAR by configuration

soulgalore:


tobli:
👍

tobli:
Could we add a test case for this?

soulgalore:
yes let me see, I'll fix PerfCascade first we can show the HTML there.

Issue #261, opened 2017-02-07
_was_pushed in chrome is inconsistent


worenga:
Currently, i am performing measurements with pages that use HTTP/2 Server push.

However, i found that the "_was_pushed" key in the har of browsertime is inaccurate or misleading.
Note for all Websites presented here i am not affiliated with any of them. They are just examples and came up in a scan. 
I stumbled upon a multitude of anomalies here are some examples:

1. ) Example Website: https://erisrenee.com/ 

This website pushes 3 resources:

```
nghttp https://erisrenee.com -v | grep -C4 PUSH
[  0.601] recv (stream_id=13) :path: /assets/css/screen.css?v=84ca1c7d60
[  0.601] recv (stream_id=13) :scheme: https
[  0.601] recv (stream_id=13, sensitive) accept-encoding: gzip, deflate
[  0.601] recv (stream_id=13, sensitive) user-agent: nghttp2/1.14.0-DEV
[  0.601] recv PUSH_PROMISE frame <length=74, flags=0x04, stream_id=13>
          ; END_HEADERS
          (padlen=0, promised_stream_id=2)
[  0.601] recv (stream_id=13) :authority: erisrenee.com
[  0.601] recv (stream_id=13) :method: GET
[  0.601] recv (stream_id=13) :path: /assets/css/github.css?v=84ca1c7d60
[  0.601] recv (stream_id=13) :scheme: https
[  0.601] recv (stream_id=13, sensitive) accept-encoding: gzip, deflate
[  0.601] recv (stream_id=13, sensitive) user-agent: nghttp2/1.14.0-DEV
[  0.601] recv PUSH_PROMISE frame <length=74, flags=0x04, stream_id=13>
          ; END_HEADERS
          (padlen=0, promised_stream_id=4)
[  0.601] recv (stream_id=13) :authority: erisrenee.com
[  0.601] recv (stream_id=13) :method: GET
[  0.601] recv (stream_id=13) :path: /assets/js/index.js?v=84ca1c7d60
[  0.601] recv (stream_id=13) :scheme: https
[  0.601] recv (stream_id=13, sensitive) accept-encoding: gzip, deflate
[  0.601] recv (stream_id=13, sensitive) user-agent: nghttp2/1.14.0-DEV
[  0.601] recv PUSH_PROMISE frame <length=73, flags=0x04, stream_id=13>
```
`/assets/css/screen.css?v=84ca1c7d60`,  
[...]

PR #265, opened 2017-02-11
Workaround for sitespeedio/browsertime/issues/261

worenga:


tobli:
I'm reluctant to adding a workaround until the core issue is fully understood. If I've understood #261 correctly, this would just catch pushed resources in the cases where they're pushed more than once, correct?

worenga:
In part, from my experiments it seems in case of a push `entry._servedFromCache` is sometimes true. Even though the browser should not have this resources cached. I am not familiar with the chrome internals but i assume the push is stored in some kind intermediary cache. 

That said, you are right. If the pushed content was actually cached this fix will obfuscate this fact.

tobli:
Closing this, since we opted for #267 instead.

PR #267, opened 2017-02-11
Improved detection of h2 pushed assets for Chrome.

tobli:


tobli:
This is similar to #265, but instead detects when resources are put in cache just between when they've been pushed, and when the parser detects and requests them.

tobli:
Let's wait for feedback from @worenga in #261 before merging this.

worenga:
I think it is ok to merge for now, but it doesn't entirely fix the issue, since the perflog is incomplete with regard to pushed resources. So yes, it is in improvement as more pushed streams are found but it does not solve the core issue.

Issue #268, opened 2017-02-12
Ability to set Firefox Preferences


worenga:
Similar to `--chrome.args` we should add a command line option to allow advanced users to specify custom firefox preferences such as `network.http.spdy.enabled.http2`. Surely a user can hack 
https://github.com/sitespeedio/browsertime/blob/master/lib/core/webdriver/firefox.js#L7
for her needs, but it still inflexible when trying multiple configurations.

soulgalore:
Good idea, let me fix that tonight.

tobli:
This will be in the next release! 🎉

PR #269, opened 2017-02-12
Set firefox preference through the CLI #268

soulgalore:


lib/core/webdriver/firefox.js
```diff
@@ -211,6 +211,14 @@ module.exports.configureBuilder = function(builder, options) {
     profile.setPreference(pref, defaultFirefoxPreferences[pref]);
   });
 
+  if (options.firefox && options.firefox.preference) {
```
tobli: Move this down, below ```const firefoxConfig = options.firefox || {};```

lib/core/webdriver/firefox.js
```diff
@@ -211,6 +211,14 @@ module.exports.configureBuilder = function(builder, options) {
...
+    const userPrefs = Array.isArray(options.firefox.preference) ? options.firefox.preference : [options.firefox.preference];
```
tobli: Could you extract ```toArray```from engine.js to a util model, and use that here too? Then you can skip the if statement altogether, just
```for (const pref of toArray(firefoxConfig.preference))```

File: lib/core/webdriver/firefox.js [hunk 3]
```diff
@@ -211,6 +211,14 @@ module.exports.configureBuilder = function(builder, options) {
...
+    for (const pref of userPrefs) {
+        const nameAndValue = pref.split(':');
+        profile.setPreference(nameAndValue[0], nameAndValue[1]);
```
tobli: Should we have a check that nameAndValue has length 2?

lib/core/webdriver/firefox.js
soulgalore: adding just before so we can override all values but still all prefreences on the same place.

lib/core/webdriver/firefox.js
soulgalore: yep, let me fix that.

lib/core/webdriver/firefox.js
soulgalore: I'll move that.

lib/core/webdriver/firefox.js
```diff
@@ -242,6 +236,18 @@ module.exports.configureBuilder = function(builder, options) {
...
+        const nameAndValue = pref.split(':');
+        if (nameAndValue.length === 2) {
+          profile.setPreference(nameAndValue[0], nameAndValue[1]);
+        } else {
+          log.error('Firefox preferences %s need to of the format key:value, preference was not set', pref)
+        }
+    }
+  }
+
   const firefoxConfig = options.firefox || {};
```
tobli: If you move this just above the if statement, there's no need for ```if (options.firefox```. Also util.toArray handles undefined, to you can skip the if altogether and just have ```const userPrefs = util.toArray(firefoxConfig.preference);```

PR #274, opened 2017-02-17
Added --experimental.dumpChromeNetlog

worenga:


worenga:
There seems to be an issue that the log is not entirely written and truncated.

lib/core/engine.js
```diff
@@ -124,6 +124,15 @@ class Engine {
       options.postScript.unshift(...videoPostScripts);
     }
 
+    if(options.experimental.dumpChromeNetlog) {
```
soulgalore: Can you change that to use lodash.get or add an extra if, yargs doesn't populate experimental so it will try .dumpChromeNetlog on an undefined object if you don't send any experimental arguments. same in chrome.js.

soulgalore:
> There seems to be an issue that the log is not entirely written and truncated.

Ok so it's unusable?

worenga:
Yes, it seems to be an issue with the chromedriver, waiting for replies here: https://groups.google.com/forum/#!topic/chromedriver-users/T5uxtJ3kKuU

worenga:
Seems to be an issue in Chromedriver: 
https://bugs.chromium.org/p/chromedriver/issues/detail?id=1688&q=&sort=-id&colspec=ID%20Status%20Pri%20Owner%20Summary

soulgalore:
Maybe we can fix it temporary by sending all the default settings to Chromedriver as the answer in your bugreport just so you get the info you need?

worenga:
there is now a fix available, i guess it will land in chromedriver 2.28. There still seems to be an related issue with windows, but im not sure if that is concerning us at all. If i manage to find some time im gonna recompile chromium and test the patch it with browsertime over the weekend

soulgalore:
@worenga we use the 2.28 in master, do you have time to test if it works?

worenga:
yes, the log ins no longer truncated in the new chromedriver.

tobli:
🎉 🎉 🎉 
Merging!

Issue #262, opened 2017-02-08
--chrome.args='--start-maximized' has no effect


worenga:
I am trying to perform fullscreen measurements with the `--start-maximized` flag for chrome.
However, i found that this flag is ignored. 

I used
```
./bin/browsertime.js -n 1 --chrome.args="--start-maximized"  https://github.com/sitespeedio/browsertime -n1 -vvv
```
for testing.
I have tried Chrome 54, Chrome 57 and Chromium (58) master.

tobli:
Interesting, is there a specific reason you'd specifically want to test full screen? Are you expecting different performance characteristics? I'd assume the viewPort is more reliable to force a specific size (to activate different mediaQuery results). Viewport uses native Selenium commands, so it'll work across different browsers. That said I'm not sure --start-maximized works, I didn't do anything when I ran Chrome stand alone with that flag. ```--kiosk```however did start Chrome in full screen. That said, Browsertime seems to switch back to windowed mode when adding ```--chrome.args="--kiosk"```. Not sure why yet. It'd be awesome if you'd like to dive deeper and figure out why.

worenga:
Currently, browsertime/selenium starts chrome in some kind of intermediary size and the window does not fit the screen, which is IMHO not representative when performing measurements. (Who uses such a weird intermediary resolution?)
Maybe it's os-specific. I am using Ubuntu Linux 16.04 Mate on my test clients.
When invoking chrome standalone from commandline with `--start-maximized`.
That said I overlooked the `--viewPort` option.

worenga:
Ah. viewPort (https://github.com/sitespeedio/browsertime/blob/master/lib/core/seleniumRunner.js#L62)  is always set (https://github.com/sitespeedio/browsertime/blob/master/lib/core/engine.js#L54) and defaults to 1200x960, which in my case is not maximized.

tobli:
Cool! Would it be ok to close this issue then (with --viewPort meeting your needs)? We try to stick to solutions using Selenium, to avoid issues that are specific to one browser and/or OS. --chrome.args and friends are there for advanced users, but not part of what we regularly test or maintain.

worenga:
Works for me. Maybe i'll add a PR so that one can say `--viewPort="maximized"` or so.

tobli:
👍

PR #275, opened 2017-02-17
Allow --viewPort maximize for non-xvfb runs, complements sitespeedio/…

worenga:
complements sitespeedio/browsertime#262


tobli:
Looks legit! Thanks, and sorry for the delay.

PR #276, opened 2017-02-19
fix permissions of browsertime-results in docker

worenga:
Executing the docker image as advertised on the Readme

```
$ docker build -t sitespeedio/browsertime .
$ docker run --privileged --shm-size=1g --rm -v "$(pwd)":/browsertime-results sitespeedio/browsertime -n 1 -c cable --video --speedIndex https://www.sitespeed.io/
```

will create a directory in the current pwd called www.sitespeed.io.
However, the permissions are the permissions of docker root. Hence, if one tries to delete it without sudo permission is denied. This is a little cumbersome if one wants to postprocess the generated files afterwards.

This PR uses the fs-permission of the host.

### DIFF
--- a/docker/scripts/start.sh
+++ b/docker/scripts/start.sh
@@ -26,3 +26,8 @@ PID=$!
 
 trap shutdown SIGTERM SIGINT
 wait $PID
+
+RESULTDIR="/browsertime-results"
+if [ -d $RESULTDIR ]; then 
+  chown -R `stat -c "%u:%g" ${RESULTDIR}` ${RESULTDIR}
+fi


beenanner:
Hmm looks like this only works if you change the owner to the host user for browsertime-results?

worenga:
Not sure i get your question. But it will change the ownership of the subfolders and contents inside the browsertime-results directory to whomever owns that directory i.e. has mounted it.

beenanner:
Sorry was just calling out that this works, but only if the folder preexist on the host machine with the correct owner. Which is probably good enough for your use case. :-)

soulgalore:
we need to fix one more thing, you can set the name of the output folder, so that needs to be take care of.

worenga:
I do not know
a) How to parse the --output flag within the shell script, and 
b) in case the folder is not already present where to obtain the host permissions from.

soulgalore:
yes I know maybe we should do this inside nodejs. However I think this should be common Docker problem? So I'm thinking maybe there's a Docker solution out there? I wouldn't mind adding a new user inside of Docker if there's a good pattern of making sure it can write outside the container.

beenanner:
If we can create a non-root user in the image then you should be able to mount the existing user like so

``` 
--user sitespeedio
```

This would require a "sitespeedio" user be created outside of docker on the host machine to match though. Passing the existing user id isn't very flexible either, since it would require that the user created inside the container have the same uid as the user passed in.

```
--user `id -u`
```

If we can setup the user in nodejs in the startup script that would be the most flexible but would have to make sure the new user has permissions to all the commands to execute. I'm all for this if possible too as well as getting rid of --privileged and passing the specific caps it needs as a best practice.

soulgalore:
I think there's other ways to do this, lets open an issue first and re-work this there. See https://github.com/sitespeedio/browsertime/issues/307

Issue #271, opened 2017-02-14
tc and tsproxy shaping differ significantly


worenga:
When comparing the load times of between tc and tsproxy i've noticed that they differ quite a lot

First some baseline measurement for `https://google.de`
```
➜  browsertime git:(261-detect-push) ✗ ./bin/browsertime.js https://google.de                                                         
[2017-02-14 11:47:05] Running chrome for url: https://google.de
[2017-02-14 11:47:05] Testing url https://google.de run 1
[2017-02-14 11:47:10] Testing url https://google.de run 2
[2017-02-14 11:47:14] Testing url https://google.de run 3
[2017-02-14 11:47:19] 21 requests, 752.09 kb, firstPaint: 670ms (±19.25ms), DOMContentLoaded: 725ms (±20.83ms), Load: 1.75s (±22.66ms), rumSpeedIndex: 762 (±9.68) (3 runs)
[2017-02-14 11:47:19] Wrote data to browsertime-results/google.de/2017-02-14T114704+0100
```

Now use tc shaping with cable and 3g profile

```
➜  browsertime git:(261-detect-push) ✗ ./bin/browsertime.js --connectivity.engine tc --connectivity.profile cable https://google.de --connectivity.tc.device enp0s25
[2017-02-14 11:36:43] Running chrome for url: https://google.de
[2017-02-14 11:36:43] sudo tc qdisc add dev enp0s25 root netem delay 28ms loss 0% rate 5000kbps
[2017-02-14 11:36:43] Testing url https://google.de run 1
[2017-02-14 11:36:49] Testing url https://google.de run 2
[2017-02-14 11:36:55] Testing url https://google.de run 3
[2017-02-14 11:37:00] 21 requests, 752.13 kb, firstPaint: 898ms (±5.54ms), DOMContentLoaded: 959ms (±8.02ms), Load: 2.23s (±31.68ms), rumSpeedIndex: 1061 (±14.12) (3 runs)
```
```
[2017-02-14 11:37:00] Wrote data to browsertime-results/google.de/2017-02-14T113643+0100
[2017-02-14 11:37:00] sudo tc qdisc del dev enp0s25 root
➜  browsertime git:(261-detect-push) ✗ ./bin/browsertime.js --connectivity.engine tc --connectivity.profile 3g https://google.de --connectivity.tc.device enp0s25
[2017-02-14 11:37:18] Running chrome for url: https://google.de
[...]

PR #277, opened 2017-02-21
Connectivity engine throttled outside of Browsertime

soulgalore:


worenga:
1. ) I am not sure why this is necessary. When profile is native neither tc nor tsproxy are started?
2. ) Also throttled might be a confusing name for essentially a no-op.

soulgalore:
1) It seems like the best way to throttle a connection for Docker is to do it on a bridge (https://github.com/sitespeedio/browsertime/issues/271#issuecomment-281171534) so we need to that outside the container. There could be other ways of fixing this in the future but one that at least would work now is the bridge solution.

2) Yes a better name sounds good, let see if @tobli and @beenanner has any suggestions.

tobli:
We're not currently picking up connectivity profile/name from Browsertime when creating Graphite metric paths, are we? That'd be a great use case for this. Also storing throttling settings in the Browsertime result json, when run stand alone (to remember the settings used when doing the measurement). One option to name a specific set of conditions (either throttled, or run e.g. via a tethered phone) would be to call the engine "named", to indicate that it's a given set of network conditions that we want to associate with a name.

soulgalore:
> We're not currently picking up connectivity profile/name from Browsertime when creating Graphite metric paths, are we?

No but since it is the same API it's clean to have it implemented in BT and then just add the CLI in sitespeed.io. 

Agree we should have it in the JSON output in BT. 

Naming is hard :)

PR #286, opened 2017-03-06
Save To mongoDB 

lerayj:
**Save HAR file result on MongoDB.**

Two new options added: 
--mongo.URI
--mongo.collection

**Why do I've done this?**

For my use case, I needed to store the HAR result file elsewhere than in the filesystem. When using stateless Docker, I cannot rely on this storage strategy. JSON type files are trivial to save on MongoDB, that's why I wrote this connector.

I can rework on this PR,  feel free to ask me change or add things on it.

Hope this could be useful to someone, I dunno if this change make sense to you guys!

### DIFF
--- a/bin/browsertime.js
+++ b/bin/browsertime.js
@@ -6,6 +6,7 @@ let Engine = require('../').Engine,
   logging = require('../').logging,
   cli = require('../lib/support/cli'),
   StorageManager = require('../lib/support/storageManager'),
+  DatabaseManager = require('../lib/support/databaseManager'),
   Promise = require('bluebird'),
   merge = require('lodash.merge'),
   isEmpty = require('lodash.isempty'),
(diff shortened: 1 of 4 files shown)


soulgalore:
Hey @lerayj, wow this is cool! 

I've talked with @tobli and @beenanner and we all agree that we wanna keep Browsertime as lean as possible, so it is easy for us to maintain and easy to re-use with as little overhead as possible.

However, this is still super cool and usable for more than you, what do you think about using Browsertime as a NodeJS library (as we do in [sitespeed.io](https://github.com/sitespeedio/sitespeed.io/tree/master/lib/plugins/browsertime) and the [Coach](https://github.com/sitespeedio/coach/blob/master/lib/index.js) and then add the capability to store the data to MongoDB there? We would happily add a link to your repo, so people who need that functionality finds it from Browsertime? Also if we need to do any changes to make that repo easier for you, just let us know and will try to fix that!

Best
Peter

lerayj:
Sure, I'm gonna look forward to implement this for the Node library!

I understand  you need to keep things clean and lean, would be happy to be mention there.

I probably gonna do a PR to add some documentation about how to use browsertime as a Node module (this is not really clear yet, you need to look at the code to use it) if you don't see any inconvenience.

tobli:
Thanks for the PR @lerayj! Based on the discussion above, I'm closing this now.

PR #290, opened 2017-03-23
Set logger name

soulgalore:


soulgalore:
We don't need this right now

PR #291, opened 2017-03-23
rename experimental and dump to chrome and collect

soulgalore:


tobli:
Should we perhaps parse the old names for a while, print a warning a convert to the new?

lib/support/cli.js
```diff
@@ -117,16 +117,26 @@ module.exports.parseCommandLine = function parseCommandLine() {
...
-      describe: 'Dump Chromes traceCategories log to disk.',
+      describe: 'Collect Chromes traceCategories log to disk.',
       group: 'chrome'
     })
     .option('chrome.traceCategories', {
       describe: 'Set the trace categories.',
       type: 'string',
       group: 'chrome'
     })
+    .option('chrome.collectChromePerflog', {
```
tobli: Now that it's in the Chrome category, I suggest we shorten it to chrome.collectPerflog.

lib/support/cli.js
```diff
@@ -117,16 +117,26 @@ module.exports.parseCommandLine = function parseCommandLine() {
...
+      type: 'boolean',
+      describe: 'Collect Chromes performance log with Network.* events and save to disk.',
+      group: 'chrome'
+    })
+    .option('chrome.collectChromeNetlog', {
```
tobli: Same as above, chrome.collectNetlog

lib/support/cli.js
```diff
@@ -117,16 +117,26 @@ module.exports.parseCommandLine = function parseCommandLine() {
       describe: 'Choose which device to use. If you do not set it, random device will be used.',
       group: 'chrome'
     })
-    .option('chrome.dumpTraceCategoriesLog', {
+    .option('chrome.collectTraceCategoriesLog', {
```
tobli: Either use a lower case L for log, or use upper case for Perflog and Netlog.

lib/support/cli.js
```diff
@@ -117,16 +117,26 @@ module.exports.parseCommandLine = function parseCommandLine() {
...
       type: 'boolean',
-      describe: 'Dump Chromes traceCategories log to disk.',
+      describe: 'Collect Chromes traceCategories log to disk.',
```
tobli: My understanding is that this means "include trace events in the perfLog". I suggest renaming the option and description to something that reflects that. Also, should this imply chrome.collectPerfLog? Maybe also chrome.traceCategories should imply this option (and by extension chrome.collectPerfLog).

soulgalore:
Taking PRs for the PR!

tobli:
If it's ok with you I can push to the PR branch as well.

soulgalore:
yes please that will make it much faster, it's so rare that we disagree :)

PR #293, opened 2017-03-26
Create output directory before starting a run.

tobli:
Also ensure output directory is created before starting a run (and before storageManager is passed to preScripts).


tobli:
ping @worenga - this is to address https://github.com/sitespeedio/browsertime/commit/cfde5a1fefd98bb2f89462dcad53cb48e7b5da13#commitcomment-21490238

tobli:
Hmm, I totally misread this. We already skip the timestamp directory when --resultDir is used. The part about ensuring output directory is created is still valid. I'll update the PR.

PR #299, opened 2017-03-29
[DON't MERGE] Run Travis tests inside a Docker container.

tobli:
Include test files and eslint config in base image for now, might revisit that later. Then build separate test container after that, to install dev dependencies.

A better solution might ultimately be to replicate everything in the sitespeedio/browsertime Dockerfile inside the test Dockerfile.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,6 +4,8 @@
 !package.json
 !docker
 !vendor
+!.eslint*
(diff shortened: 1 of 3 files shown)


soulgalore:
We have a new setup now when we run tests on different OS so let us close this.

PR #301, opened 2017-03-31
External is default connectivity and remove ts proxy until it works

soulgalore:


soulgalore:
@tobli also wants your approval before I push this one, so we can move on with 1.0

PR #315, opened 2017-04-18
Patch Selenium in Docker for Firefox extensions

soulgalore:
I really want to get 1.1 out so we can release sitespeed.io 5.0 and it kind of sad that Firefox will miss support blocking and adding request headers.  I think we either should patch Selenium and run our own repo fo a release or two or hack it in Docker as in this repo.


Dockerfile
```diff
@@ -16,5 +16,7 @@ COPY . /usr/src/app
 WORKDIR /
 
 COPY docker/scripts/start.sh /start.sh
+# Patch until https://github.com/SeleniumHQ/selenium/pull/3846 gets released
+COPY docker/selenium/extension.js /usr/src/app/node_modules/selenium-webdriver/firefox/extension.js
```
kkopachev: is this file missing from the PR?

soulgalore: ooopps yes, pushing it now.

soulgalore:
This was finally released a month ago.

Issue #316, opened 2017-04-19
Firefox 53 fails 


soulgalore:
There's a current miss-match between releases of Firefox 53, Geckodriver and Selenium (NodeJS version). If you run the current Browsertime with FF 53 you get:
<pre>
[2017-04-19 21:14:03] Error running browsertime BrowserError: Unrecognised timeout: page load
</pre>

When Selenium 3.4.0 and Geckodriver 0.16.0 is released, we can upgrade and fix that. Keep using our Docker containers :)

tobli:
Fail no more! Works as of 06b9d35.

PR #317, opened 2017-04-20
Catch pageLoadTimeout errors so that current Selenium work with FF53 …

soulgalore:
See #316


tobli:
Not needed now since the selenium 3.4 upgrade! 🎉

Issue #312, opened 2017-04-18
Run Browsertime on Android phones via Docker


soulgalore:
If you install Browsertime using npm you can attach your Android phone following [this setup](https://www.sitespeed.io/documentation/sitespeed.io/mobile-phones/). Then you can collect metrics, HAR file and if you install the dependencies for VisualMetrics you can also collect a video and get SpeedIndex and other related visual metrics.

It would be super if we could make the same work in our Docker container (then you don't need to install all the dependencies needed for VisualMetrics). 

The current version of our Docker container doesn't add the libraries needed for adb, so that needs to be fixed. I haven't find away to mount the usb ports on Mac to Docker but it "should" work on Ubuntu.

We need to:
* Add the libraries needed for running ADB
* Test and add documentation how to run on Ubuntu using the Docker container.

What we can look into these links to get some inspiration:
https://github.com/sorccu/docker-adb
https://groups.google.com/forum/#!msg/docker-user/UsekCwA1CSI/RtgmyJOsRtIJ

soulgalore:
Fixed in https://github.com/sitespeedio/browsertime/commit/aab16ec0b8610e3e4bc23798ee181ffb9a14953d

PR #322, opened 2017-04-27
[DO-NOT-MERGE] Run adb from Docker

soulgalore:
We wanna shuffle the code and start the adb server automatically, so we should change this before we merge.

This works now for me on Ubuntu 16, so please try it @beenanner when you have time!
Build the same as before:
`sudo docker build -t sitespeedio/browsertime .`

And then run (you need to run without  XVFB to make it work):
`docker run --privileged -v /dev/bus/usb:/dev/bus/usb --shm-size=1g --rm -v “$(pwd)“:/browsertime-results sitespeedio/browsertime -n 1 --chrome.android.package com.android.chrome --xvfb false --speedIndex --video https://www.sitespeed.io`

If we can verify that this works for others I can fix the rest :)

See #312 

Ping @abhagupta (need to run Docker on Linux).


soulgalore:
We got final version coming ... now!

PR #325, opened 2017-05-03
support for saucelabs, setting custom capabilities

abhagupta:
@soulgalore This PR is in reference to https://github.com/sitespeedio/sitespeed.io/issues/1393 . Had to do a small change to get selenium url correctly. 

Tests Passed locally.

 Found one limitation, Saucelabs looks for a capability `--version` in order to open a specific version of browser if passed in by user. For example, you can open `chrome 56` specifically.  If not present, it opens default.   However, if `--version` is passed in here, `browsertime` detects it as a query on its own version. I have not addressed this issue in this PR.  Please let me know if there is any suggestion around this.  

Also please feel free to put the doc anywhere else where it makes more sense.


soulgalore:
Aha we pass all options to Selenium today (I don't remember)?. Thinking maybe we can add one extra layer so we have <code>--saucelabs.version</code> <code>--saucelabs.platform</code> or will it mess up? Need to register an account and test and discuss with @tobli and @beenanner!

soulgalore:
It will take a while, need to digest https://saucelabs.com/terms-of-service first :)

abhagupta:
Yep :). I have a corporate account so it was easy for me.

abhagupta:
@soulgalore  this PR has been updated to support any cloud service.  Most of these services just have a custom capabilities and if browsertime can set that for user, we should be good. Please let us know if this PR will be considered. Because it is a must-have for us.

soulgalore:
Thanks @abhagupta let me have a look asap, latest tonight. ping @tobli

soulgalore:
Sorry for not getting back. Been discussing it with @tobli during the day. I think we both are scared that this would open Pandora's box, making it hard for us to make sure all aspects of Browsertime will work. One way could be to be a simple implementation as a plugin for sitespeed.io running browsers at third party providors. I'll get back tomorrow again with pro/cons and how we are thinking (@tobli feel free to add your ideas).

abhagupta:
Agreed with your concerns @soulgalore and @tobli . May be the PR should just be thought of as an ability to set your custom capabilities, rather than connecting to third party vendor. Custom capabilities can be anything from running it on local mobile device, tablet, or cloud to enabling tracing, perf logging, different version of chrome etc. etc.    Also, if adding custom capabilities causes an issue in the execution of sitespeed, it becomes a responsibility of the user, not sitespeed.io.  IMO, sitespeed  project will become even more usable and flexible if you let users set custom capabilites, just like webdriver.io.  Infact, most of the projects which use selenium-webdriver let you set custom capabilities.

 Btw, I have been running sitespeed.io with all kinds of custom capabilities, like running with Appium, different chrome versions, different cloud services, and its working like a charm as long as I don't mess with anything else in browsertime core.

Hope this helps :)

abhagupta:
@soulgalore  Anything on this?

soulgalore:
Let me sit down tonight.

soulgalore:
Hi @abhagupta yep I agree, let us keep just where we set the capabilities. We can remove the docs/test for now and keep it a secret for those who knows and let it mature for a while. If you have time to do the fix, I can do a release of BT tonight and sitespeed.io too. Else I can just cherry pick and fix it tonight so we have a shiny new release for Monday.

Best
Peter

soulgalore:
Also we can rename the option to `--selenium.capabilities ` !

abhagupta:
thanks @soulgalore  I will do it in couple of hours

abhagupta:
@soulgalore  PR updated. Let me know if there is any other change required

soulgalore:
thanks @abhagupta btw would love to see a blog post about all the creative usage you've done with sitespeed/browsertime!

abhagupta:
absolutely. Already on it.  I have some excellent examples of sitespeed used on appium, saucelabs, testobject, browserstack etc.    Thanks for merging.

PR #332, opened 2017-05-11
Bring the netlog from Android

soulgalore:


lib/support/video/scripts/pullNetLogAndroid.js
```diff
@@ -0,0 +1,23 @@
+'use strict';
+
+const Promise = require('bluebird');
+const fs = require('fs');
+
+module.exports = {
+    run(context) {
+        const taskData = context.taskData;
+        context.log.debug('Pull netlog');
+        const file = `/sdcard/chromeNetlog-${ context.index}.json`
```
worenga: This looks like a closed world assumption to me. Can we be sure the netlog always appears there?

soulgalore: yes we do that when we set the chrome options, check https://github.com/sitespeedio/browsertime/pull/332/files#diff-1c3ad552ac7bae72e3768fea0e5bb02eR108

PR #339, opened 2017-05-31
Restore ability to set http(s) proxy for Chrome.

### LINKED ISSUE #338: Proxy don't work in Chrome
If I set --proxy.https/proxy.http I cannot get Chrome to use a proxy. Doing it with --chrome.args works.

tobli:
When setting ChromeOptions, any proxy settings will be overwritten (see https://github.com/SeleniumHQ/selenium/blob/d7bdd7cc42faafb1eba031c26b7c369cdea026d5/javascript/node/selenium-webdriver/chrome.js#L654). Fix by setting our proxy settings after any browser specific options.

For the same reason, there’s no need to specify browser to the webdriver.Builder. By setting ChromeOptions/FirefoxOptions the browser name will be overwritten anyway.

Fixes #338.

### DIFF
--- a/lib/core/webdriver/index.js
+++ b/lib/core/webdriver/index.js
@@ -17,15 +17,7 @@ let Promise = require('bluebird'),
 module.exports.createWebDriver = function(options) {
   const browser = options.browser || 'chrome';
   const seleniumUrl = options.selenium ? options.selenium.url : undefined;
-  const builder = new webdriver.Builder()
-    .forBrowser(browser)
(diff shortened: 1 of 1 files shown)


worenga:
should we hack around this setting the chrome.args manually if we see that a proxy is specified?
or shall we just make it a known issue?

tobli:
Ok, one more try. This time with at least some testing…

See https://github.com/sitespeedio/browsertime/pull/340

PR #344, opened 2017-06-05
Collect metrics from the paint API (coming in Chrome 60)

soulgalore:
Let us iterate what kind of structure we want of the output.


soulgalore:
If https://bugs.chromium.org/p/chromium/issues/detail?id=727046#c21 gets pushed to 60-beta, then we should upgrade the Docker containers to use that and include the paint api.

Issue #289, opened 2017-03-17
Collect browser console.log/error


soulgalore:
It would cool to collect and include it in the result JSON. Let us make it configurable so you can choose to include. We can use driver.manage().logs().get(logging.Type.BROWSER)  or something like that :)

soulgalore:
This is done for Chrome.

PR #345, opened 2017-06-09
Collect the log from Chrome and add it to the result JSON #289

soulgalore:


soulgalore:
Can you look at the format @tobli @beenanner , turn it on by --chrome.collectConsoleLog. Now we collect all different levels and keep it in the original format. 

`"consoleLog":[[{"level":"INFO","message":"https://www.aftonbladet.se/hyper-assets/preload.1d106fcef4b37d1f700d.js 2:28108 \"AST library loaded: 0.12.2\"","timestamp":1518353971623,"type":""},{"level":"WARNING","message":"https://www.aftonbladet.se/ - The SSL certificate used to load resources from https://ib.adnxs.com will be distrusted in M70. Once distrusted, users will be prevented from loading these resources. See https://g.co/chrome/symantecpkicerts for more information.","timestamp":1518353972155,"type":""}`

soulgalore:
Updated after discussion with @tobli : Store the console log to a file instead of adding the data to the browsertime.json. Much better!

lib/core/engine.js
```diff
@@ -275,6 +275,16 @@ class Engine {
             );
           }
         })
+        .tap(results => {
+          if (options.chrome && options.chrome.collectConsoleLog) {
+            runner
```
tobli: This produces a promise, so it needs a return so the outer promise chain waits for it.

PR #347, opened 2017-06-23
Upgrade to 2.30 with haxx

soulgalore:
@tobli and @beenanner can you help me verify that this really works? make sure to _rm -fR node_modules_ and re-install and run:
` bin/browsertime.js --chrome.collectTracingEvents -n 1 https://www.sitespeed.io`
for a couple of different sites.

### DIFF
--- a/lib/core/engineDelegate/chromeDelegate.js
+++ b/lib/core/engineDelegate/chromeDelegate.js
@@ -39,7 +39,11 @@ class ChromeDelegate {
     }
 
     // remove irrelevant entries from performance log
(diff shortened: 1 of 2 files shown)


soulgalore:
Even though the hack to make it work is very special, I think we should it out on DO, we need keep up to date with latest Chromedriver versions.

beenanner:
👍 sounds good I'll test a bit tonight when the wife goes to bed.

beenanner:
Ran this through a few of the CNET pages and everything seems ok @soulgalore anything specific I should be looking out for? :-)

soulgalore:
Perfect! When it's not working the size is 17kb!

PR #355, opened 2017-08-01
Fix text error about postScript

StevenUpForever:


tobli:
Thanks! Sorry for the long delay.

PR #363, opened 2017-09-08
Use browsertime as workdir 

soulgalore:
to make it possible to link in pre/post scripts and follow sitespeed.io see: https://github.com/sitespeedio/sitespeed.io/issues/1703


tobli:
Looks good!

PR #364, opened 2017-09-10
Set FPS from the cli (instead of hardcoded 60fps)

soulgalore:
Any ideas for a name of the parameter?


beenanner:
Is the idea to allow users to restrict the fps so they can manage video file size?

beenanner:
Maybe videoParams.fps to allow for additions in the future?

soulgalore:
@beenanner yes sorry I missed the reason: setting FPS can decrease the size and makes it easier to compare the visual metrics with other tools that have lower fps.

wondering if there's a way to fix the current limitations with yargs (like --video and --video.fps that don't work side by side)? else videoParams sounds good.

Issue #371, opened 2017-09-25
Silence xvfb logging 


soulgalore:
Installing on a clean Ubuntu 14 gives me the following log:
```
root@ubuntu-webpagereplay:~# browsertime --xvfb https://www.sitespeed.io -n 1 -b chrome --chrome.args no-sandbox --video --speedIndex
[2017-09-25 08:03:10] Running chrome for url: https://www.sitespeed.io
Initializing built-in extension Generic Event Extension
Initializing built-in extension SHAPE
Initializing built-in extension MIT-SHM
Initializing built-in extension XInputExtension
Initializing built-in extension XTEST
Initializing built-in extension BIG-REQUESTS
Initializing built-in extension SYNC
Initializing built-in extension XKEYBOARD
Initializing built-in extension XC-MISC
Initializing built-in extension SECURITY
Initializing built-in extension XINERAMA
Initializing built-in extension XFIXES
Initializing built-in extension RENDER
Initializing built-in extension RANDR
Initializing built-in extension COMPOSITE
Initializing built-in extension DAMAGE
Initializing built-in extension MIT-SCREEN-SAVER
Initializing built-in extension DOUBLE-BUFFER
Initializing built-in extension RECORD
Initializing built-in extension DPMS
Initializing built-in extension Present
Initializing built-in extension DRI3
Initializing built-in extension X-Resource
Initializing built-in extension XVideo
Initializing built-in extension XVideo-MotionCompensation
Initializing built-in extension SELinux
Initializing built-in extension GLX
[dix] Could not init font path element /usr/share/fonts/X11/misc, removing from list!
[2017-09-25 08:03:10] Testing url https://www.sitespeed.io run 1
6 XSELINUXs still allocated at reset
SCREEN: 0 objects of 264 bytes = 0 total bytes 0 private allocs
DEVICE: 0 objects of 96 bytes = 0 total bytes 0 private allocs
CLIENT: 0 objects of 144 bytes = 0 total bytes 0 private allocs
WINDOW: 0 objects of 48 bytes = 0 total bytes 0 private allocs
PIXMAP: 1 objects of 16 bytes = 16 total bytes 0 private allocs
GC: 4 objects of 16 bytes = 64 total bytes 0 private allocs
CURSOR: 1 objects of 8 bytes = 8 total bytes 0 private allocs
TOTAL: 6 objects, 88 bytes, 0 allocs
1 PIXMAPs still allocated at reset
[...]

PR #372, opened 2017-09-25
Silent xvfb by default. If you enable -vv or higher, you will get the logs #371

soulgalore:


soulgalore:
let me do one push ...

PR #381, opened 2017-10-15
Move out engine utils to own utils file as step 1 of making the engine smaller

soulgalore:
The engine js-file is huge, lets start to clean it up to just move out the utility methods to another file.


tobli:
👍

PR #382, opened 2017-10-15
Make 30fps default instead of 60

soulgalore:


tobli:
Thanks for all the testing, 30 seems like a better default.

PR #394, opened 2017-11-06
Use default values instead of taking undefined from cli

soulgalore:
Lets add these values as constants before we merge. Then we can remove these defaults in sitespeed.io and have cleaner code.

### DIFF
--- a/lib/support/video/finetune/ffmpegRemoveOrangeAndAddText.js
+++ b/lib/support/video/finetune/ffmpegRemoveOrangeAndAddText.js
@@ -48,7 +48,9 @@ module.exports = {
     ];
 
     // keeping old legacy videoRaw
-    const showTimer = options.videoRaw ? false : options.videoParams.addTimer;
+    const showTimer = options.videoRaw
+      ? false
+      : options.videoParams.addTimer || true;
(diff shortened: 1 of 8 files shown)


lib/support/video/finetune/ffmpegRemoveOrangeAndAddText.js
```diff
@@ -48,7 +49,9 @@ module.exports = {
     ];
 
     // keeping old legacy videoRaw
-    const showTimer = options.videoRaw ? false : options.videoParams.addTimer;
+    const showTimer = options.videoRaw
+      ? false
+      : options.videoParams.addTimer || videoDefaults.addTimer;
```
tobli: This is incorrect. showTimer will be set to true for the following case:
```
options = {
videoParams: {addTimer: false}
}
```

soulgalore: yep is a known bug but I haven't reported it :D

soulgalore: fixed in https://github.com/sitespeedio/browsertime/pull/396

Issue #401, opened 2017-11-16
window.setSize does not work on Android.


abhagupta:
Hi @peter  I am running sitespeed tests  on android real device via connecting to Appium. But there is a piece of code in which the test fails because of window.setSize here : https://github.com/sitespeedio/browsertime/blob/master/lib/core/seleniumRunner.js#L83. Reason:  Android does not support `window.setSize`.  Can I propose to add an option of boolean `mobile` in the options sent to sitespeed.io which determines whether this code will be executed or not.  I can send a PR.

soulgalore:
Hi @abhagupta yes please send a PR. We have a fix for it in sitespeed.io  see https://github.com/sitespeedio/sitespeed.io/blob/master/lib/plugins/browsertime/index.js#L78-L85 but we should solve it better in Browsertime. So master is hot now (coming 2.0). If you need the fix now, we need to merge in 1.x.

Best
Peter

soulgalore:
There's a hack now since 3.x. Add `--android` as a cli param and it will work.

PR #403, opened 2017-11-17
added feature for not running window.setSize when option mobile is provided

abhagupta:
PR for https://github.com/sitespeedio/browsertime/issues/401
@soulgalore @tobli

### DIFF
--- a/lib/core/seleniumRunner.js
+++ b/lib/core/seleniumRunner.js
@@ -62,26 +62,29 @@ class SeleniumRunner {
         });
       })
       .tap(() => {
-        let viewPort = this.options.viewPort;
-        if (viewPort) {
-          if (viewPort !== 'maximize') {
-            viewPort = viewPort.split('x');
-          }
-          const window = this.driver.manage().window();
-
-          if (viewPort === 'maximize') {
-            if (this.options.xvfb) {
-              log.info(
-                'Maximizing window in XVFB may not work, make sure that you verify that it works.'
-              );
+        let isChromeRunningOnMobile = this.options.mobile;
+        if (!isChromeRunningOnMobile) {
+          let viewPort = this.options.viewPort;
+          if (viewPort) {
+            if (viewPort !== 'maximize') {
+              viewPort = viewPort.split('x');
             }
-            return window.maximize();
-          } else {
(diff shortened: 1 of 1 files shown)


lib/core/seleniumRunner.js
```diff
@@ -62,26 +62,29 @@ class SeleniumRunner {
         });
       })
       .tap(() => {
-        let viewPort = this.options.viewPort;
-        if (viewPort) {
-          if (viewPort !== 'maximize') {
-            viewPort = viewPort.split('x');
-          }
-          const window = this.driver.manage().window();
+        let isChromeRunningOnMobile = this.options.mobile;
```
soulgalore: Sorry the code hard to read. The right cli option is **chrome.android.package** then we know we are on Android (you cannot run on Android without or ... do you have special hack, maybe it is possible?). You can use lodash.get to get it (we are inconsistent in the code, but that is the right way 2 do it).

soulgalore:
@abhagupta if we release 2.0 next week will you pick that up or wait to December (meaning we should release next 1.x this weekend you can use it next week?)?

abhagupta:
HI @soulgalore  Yes, I am running these tests on saucelabs devices which connect via appium url.  In this situation none of these are available :   
   ```this.options.chrome &&
      this.options.chrome.android &&
      this.options.chrome.android.package ```.    I will keep looking if I can find other way to avoid addition of this code, but I don't think the check for above option values will help us.

soulgalore:
Ok, so how do you set it up? You need to write  that blog post someday so we know how it works :)

abhagupta:
Thanks @soulgalore  for getting back.  https://medium.com/@abhaguptatweet/run-sitespeed-io-on-android-devices-4777b405faaf    - still polishing but should give you details on how to run the test. Please feel free to provide feedback/mistakes/suggestions on blog as well.

abhagupta:
@soulgalore  I dont need this change anymore.  we have extended the webdriver and overridden the `setPosition` and `setSize` functions.   I will close this PR

soulgalore:
@abhagupta aha ok cool. thanks for sharing the post, looking forward to it!

PR #405, opened 2017-11-18
Bump to latest prettier and reformat.

tobli:


lib/core/webdriver/chrome.js
```diff
@@ -47,7 +47,9 @@ const defaultTimelineTraceCategories =
 function migrateLegacyOption(options, from, to) {
   if (get(options, from)) {
     log.warn(
-      `The option ${from} has been renamed to ${to}. The old option will be removed in a future version.`
+      `The option ${from} has been renamed to ${
+        to
```
soulgalore: This looks ugly right?

tobli: Weeeell, it's not a file we change that often. =)
Most other things are pretty, and it's automatic and consistent. We'll take a few losses once in a while.

PR #407, opened 2017-11-20
Start using async/await instead of promise chains.

tobli:


soulgalore:
I'm thinking we should merge this right after 2.0/6.0 (early next week hehe)? Then we can keep it running on the dashboard for a while and minimise the risk we introduce a new bug.

PR #410, opened 2017-12-05
Docker file with WebPageReplay

soulgalore:
This follow the same pattern as https://github.com/soulgalore/browsertime-replays/tree/master/webpagereplay I've been using at work.

It works like this:
1. The start script starts WebPageReplay in record mode
2. Then starts Browsertime accessing the URL you choose one time (so it is recorded)
3. WebPageReplay is closed down
4. WebPageReplay in replay mode is started
5. Browsertime access the URL so many times you choose.

Build it:
`docker build -t sitespeedio/browsertime . -f Dockerfile.wpr`

And run (you can change the latency and number of runs with LATENCY and RUNS).

`docker run --cap-add=NET_ADMIN --shm-size=1g --rm -v "$(pwd)":/browsertime -e REPLAY=true -e RUNS=11 -e LATENCY=100 sitespeedio/browsertime https://en.wikipedia.org/wiki/Barack_Obama`

Right now we pass on all parameters to both runs of Browsertime, that is not optimal, for example we will have two result dirs (first run with no valuable metrics) and the main run(s). Any ideas how we can make the setup better?

I've updated the README. Let us keep two different Dockerfiles until FF 57 supports HAR.


docker/scripts/wpr/start.sh
```diff
@@ -0,0 +1,21 @@
...
+google-chrome --version
+
+echo 'Use Browsertime with WebPageReplay. Chrome only at the moment'
+
+RUNS="${RUNS:-5}"
+LATENCY=${LATENCY:-100}
+
+webpagereplaywrapper record --start --path /root/go/src/github.com/catapult-project/catapult/web_page_replay_go
+
+/usr/src/app/bin/browsertime.js --chrome.args host-resolver-rules="MAP *:80 127.0.0.1:8080,MAP *:443 127.0.0.1:8081,EXCLUDE localhost" -n 1 --pageCompleteCheck "return true;" "$@"
```
beenanner: Since this one is a throw away maybe just write this to /tmp with a hardcoded `--resultDir /tmp `?

soulgalore: yes but if you add resultDir in the input when you run it (you probably want to have it in a specific dir so you can do something with the metrics) we will have an array of dirs.

soulgalore: the same for example if you add -n 5 ... it's not a perfect solution right now.

beenanner: Ah I see now! I'll try some stuff out and see if I find any other ways.

soulgalore:
I think this is finally ready for testing!

PR #411, opened 2017-12-12
New default check avoiding using date

soulgalore:


soulgalore:
Ah sorry I missed the context: When we do the testing in WebPageReplay there's a hack that make sure that the Javascript Date now never change (so that each iteration will be the same). That hack makes our current pageComplete check unusable (it never times out). 

If we gonna merge the two Dockerfiles into one (we have one default now and one that includes WebPageReplay) it would be neat if we can use the same script that works across the board. However right now we cannot merge it because WPR adds something like 1gb to the container.

Let me update the script.

marty90:
BTW: I'm using the standard `browsertime` container and `WPR`. I had the issued mentioned above (the `Date` hack made `browsertime` hang). Using `--pageCompleteCheckInactivity` solved the issue for me ;)

marty90:
I'm actually using browsertime without any customization for WebPageReplay (for some other reasons). The default browsertime check makes browsertime hang even with the last version.

soulgalore:
Cool, yep then I see, thanks!

PR #427, opened 2018-01-16
Make it easy to only get the devtools.timeline

soulgalore:
The default trace categories we have is making Chrome unstable. Maybe we should just change the default to devtools.timeline, but that needs to happen in 3.0 if thats the case.


soulgalore:
Yep I agree we should change how it is handled today, let us do that in 3.0.

PR #431, opened 2018-01-24
Use chrome-devtools to collect cpu metrics

soulgalore:
There's one big downsize to use devtools-timeline-model. The size:

`─────────────────────────┬──────────────┬────────┐
│ name                    │ children     │ size   │
├─────────────────────────┼──────────────┼────────┤
│ alto-saxophone          │ 365          │ 17.95M │
├─────────────────────────┼──────────────┼────────┤
│ devtools-timeline-model │ 2            │ 17.59M │
├─────────────────────────┼──────────────┼────────┤
│ selenium-webdriver      │ 32           │ 6.91M  │
├─────────────────────────┼──────────────┼────────┤
│ chrome-har              │ 6            │ 5.00M  │
├─────────────────────────┼──────────────┼────────┤
│ soprano-saxophone       │ 2            │ 3.48M  │
├─────────────────────────┼──────────────┼────────┤
│ adbkit                  │ 10           │ 3.01M  │
├─────────────────────────┼──────────────┼────────┤
│ moment                  │ 0            │ 2.20M  │
├─────────────────────────┼──────────────┼────────┤
│ yargs                   │ 63           │ 0.85M  │
├─────────────────────────┼──────────────┼────────┤
│ xvfb                    │ 2            │ 0.64M  │
├─────────────────────────┼──────────────┼────────┤
│ bluebird                │ 0            │ 0.59M  │
├─────────────────────────┼──────────────┼────────┤
│ intel                   │ 13           │ 0.26M  │
├─────────────────────────┼──────────────┼────────┤
│ hasbin                  │ 1            │ 0.25M  │
├─────────────────────────┼──────────────┼────────┤
│ execa                   │ 15           │ 0.13M  │
├─────────────────────────┼──────────────┼────────┤
│ @sitespeed.io/throttle  │ 9            │ 0.13M  │
├─────────────────────────┼──────────────┼────────┤
│ mkdirp                  │ 1            │ 0.07M  │
├─────────────────────────┼──────────────┼────────┤
│ lodash.groupby          │ 0            │ 0.07M  │
├─────────────────────────┼──────────────┼────────┤
│ lodash.remove           │ 0            │ 0.06M  │
├─────────────────────────┼──────────────┼────────┤
│ lodash.merge            │ 0            │ 0.06M  │
├─────────────────────────┼──────────────┼────────┤
│ fast-stats              │ 0            │ 0.04M  │
├─────────────────────────┼──────────────┼────────┤
│ lodash.set              │ 0            │ 0.03M  │
[...]

PR #434, opened 2018-01-30
Use trace_parser.py to parse the trace log for Chrome

soulgalore:


soulgalore:
This is a much better alternative than https://github.com/sitespeedio/browsertime/pull/431

PR #454, opened 2018-02-19
Make engine easier to understand.

soulgalore:
This is a first go at making the engine logic easier to follow. Let us break it down into multiple files and add some docs to what's happening. Treat this as a first go changing the structure by first making it a little easier to follow.

Feel free to just update the PR with whatever we should change so we can merge this as a first step.


tobli:
Let's merge this as a first step. Smaller files will make it easier to reason about further changes. The main problem in the code base that this makes even more obvious (doesn't create, just highlights) is the overloading of "run".

Run is used:
- as a verb in method names. Often as the only exported method in modules that only does one thing (e.g. a pre script).
- as a noun to mean a collection of iterations (e.g. as in ChromeDelegate#onStartRun)
- as a noun to mean one iteration (just about everywhere… e.g. ```iteration.run(runIndex)``` )

I made one update to this before merging (changing the new perRun method in Iteration to perIteration), but we should really clean up this mess before finalising 3.0. I don't have a clear suggestion in mind just yet, but my thinking right now is that it'd be best to find a different term for the collection of iterations (i.e. the test of one url), and standardise on iteration for one browser invocation.

PR #457, opened 2018-02-20
Modernize JavaScript

soulgalore:
Use Navigation Timing V2 and more.


soulgalore:
We need to upgrade FF in travis for this to work and then we say goodbye to the HAR. So @beenanner @tobli let me know when you are ready :)

PR #458, opened 2018-02-20
Include WebPageReplay in the default Dockerfile

soulgalore:


soulgalore:
When we push this, we can say goodbye to FF 54 and we need to use skipHar until the new exporter is built.

PR #498, opened 2018-03-09
Change default complete check to look for 2s of inactivity

soulgalore:
and make it possible to set the size of the resource timing buffer.

We maybe should log if a web site hit the limit of 600 ... another alternative is to default just skip getting the resource timing API and then empty it.


soulgalore:
Lets use https://github.com/sitespeedio/browsertime/pull/574 instead.

PR #509, opened 2018-03-14
Add ability to take a screenshot in error case for verbose > 1

vgoncharenko:
In case when you disable screenshots and videos for test (for some reason in my case ffmpeg use a lot off resources on my VM) you maybe want to have screenshots in error case for understanding 'what happens during my test execution that produced an error'


soulgalore:
thanks @vgoncharenko for the PR, merging for upcoming 3.0

PR #512, opened 2018-03-18
Android tweaks

tobli:


lib/core/seleniumRunner.js
```diff
@@ -106,8 +106,8 @@ class SeleniumRunner {
           // Android do not support set position
           if (
             // Hack for Wallmart labs
-            !this.options.isAndroid &&
-            !get(this.options, 'chrome.android.package', false)
+            !this.options.android &&
```
tobli: What's the "--android" (or "--isAndroid") option? I don't see it in documentation, or anywhere else in the code.

soulgalore: I think you "invented" --android in your last commit :D Before we only checked if chrome.android.package is set. However the nice folks and Wallmarts labs had a PR before to not set the window size/location (that don't work on Android) when they run on Saucelabs (without setting chrome.android.package) but since we didn't merge they override the webdriver instead). But I think if we keep it like this, then we can help more users?

tobli: I’m not saying I didn’t, but then it wasn’t intentional! =)
Let’s keep the focus on this to settle on a good approach.

PR #516, opened 2018-03-20
Mark async methods with the async keyword.

tobli:
Using async makes it much easier to see what functions return Promises, which helps when reasoning about code flows.

### DIFF
--- a/bin/browsertime.js
+++ b/bin/browsertime.js
@@ -1,18 +1,18 @@
 #!/usr/bin/env node
 'use strict';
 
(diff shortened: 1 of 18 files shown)


tobli:
After seeing the Travis failure I realise that async functions will wrap Bluebird promises in native promises, so using bluebird specific methods doesn't work. A good thing in the end, to force all uses of Bluebird to be local, and not leak outside a function. I'll rework the PR and push a fix.

tobli:
Finally passing Travis. =)

PR #520, opened 2018-03-21
Simplify chrome trace log parsing.

tobli:
Use async/await instead of deep promise chains, to make code more readable.


lib/core/engine/collector.js
```diff
@@ -118,11 +119,15 @@ class Collector {
           data.extraJson[`trace-${index}.json`],
           this.storageManager.baseDir,
           index
-        ).then(timelinesArray => {
```
tobli: it's not an array, hence the renamed parameter.

PR #523, opened 2018-03-22
Don't alter the options object.

soulgalore:
Remove options.index and options.baseDir. As an extra bonus, we now create one MOZ-log and Chrome net-log per run, instead of having one large file.


tobli:
Just needs a rebase, then good to go!

PR #526, opened 2018-03-23
log early, catch early

soulgalore:


lib/core/engine/iteration.js
```diff
@@ -175,14 +175,13 @@ class Iteration {
       }
       return result;
     } catch (e) {
+      log.error(e);
       // In docker we can use a hardcore way to cleanup
       if (options.docker) {
         if (recordVideo) {
           await stop('ffmpeg');
         }
-        await stop(options.browser);
```
tobli: Is this not needed anymore? It seems not related to logging.

lib/support/stop.js
```diff
@@ -8,5 +8,5 @@ module.exports = function(processName) {
 
   log.debug('Kill all processes ' + processName);
 
-  return execa('pkill', scriptArgs).catch();
+  return execa('pkill', scriptArgs).catch(() => {});
```
tobli: This is to not handle the case when pkill exits with != 0, right? Which I guess is an acceptable error we don't want to cause browsertime to stop for.
If so I think using the reject=false option for execa is better, see https://github.com/sindresorhus/execa#reject

soulgalore: cool let me change that.

lib/core/engine/iteration.js
soulgalore: sorry too many commits in one.  I think it was never needed because we stop the browser with the driver? we should instead kill the driver if something fails.

PR #544, opened 2018-04-06
sharp 0.20.1

soulgalore:


tobli:
Yay for pre-built sharp binaries! One small step closer to not requiring a massive toolchain to compile during install. However this means one more outgoing connection (to https://github.com/lovell/sharp-libvips/releases) needed when installing. Might affect people doing builds in restricted environments, so would be good to mention in changelog.

tufandevrim:
Thanks for this.

PR #559, opened 2018-04-16
Use the chrome-trace module for trace parsing.

tobli:
One small change to the output format, event categories (e.g. ”scripting”) are not lowercase, instead of Capitalized.


soulgalore:
Woho! Let us merge this ASAP when Travis works again. Could you write a small change summary in the changelog @tobli ? Then we can push a release, upgrade sitespeed so we get it out there for more testing!

PR #585, opened 2018-05-16
Increase Chrome wait time too (1s->2s) as we did with Firefox

soulgalore:


soulgalore:
Lets do https://github.com/sitespeedio/browsertime/pull/590 instead

Issue #587, opened 2018-05-16
Install FFMPEG on Travis


soulgalore:
We have used the statics build on https://johnvansickle.com/ffmpeg/ but they are no more (or at least for now), so we should have a new way of installing FFMPEG. The same way in our Docker container.

soulgalore:
fixed yesterday

PR #588, opened 2018-05-16
Ffmpeg travis

soulgalore:


soulgalore:
#587

soulgalore:
screen shot 2018-05-16 at 4 11 42 pm
Ok, we miss the font, that automagically worked in the old version (we only specify the font on Mac).

soulgalore:
Puh, the problem is that ffmpeg isn't built with default font fallback `--enable-libfontconfig` see https://ffmpeg.org/ffmpeg-filters.html#drawtext-1

PR #597, opened 2018-05-21
let the extension server use the --android flag

soulgalore:


soulgalore:
Let is cleanup this later and move it into the android configured function.

Issue #591, opened 2018-05-17
Capture network traffic in a .pcap file


marty90:
Hi,
I think a useful feature could be to create the `.pcap` file containing the network traffic exchanged by the browser and the server, like `webpagetest` does.
It should not be difficult to implement. Just start `tcpdump` before starting the browser, and close it immediately after.

worenga:
I like the idea for pure convenience and for having the feature to further analyze page loads on a deeper level. It may help future analysis and the development of novel insights. 

However, it will introduce browsertime into the packet capture business and will create a lot of new things we may need to worry about in the future, if we support this. I am a bit worried that it is more than simply starting/stopping tcpdump. e.g. for inspecting layer 7 we may need the encryption keys used for https, browsers do handle that differently.

marty90:
Capturing the raw .pcap file should be simple. WebPageTest uses WinPcap just to read packets from an interface and dump them to a file ([here](https://github.com/WPO-Foundation/webpagetest/blob/511f3e65cff10a4a3f4d08ceea6d736a77f9253a/agent/wpthook/WinPCap.cpp#L228)). TcpDump works in the same way.
Having the raw Pcap file would be interesting to see TCP related information, or data about TLS handshake.

I agree that having the TLS keylog would be useful to decrypt the stream. That is simple for Chrome and Firefox (just set an environment variable), but for mobile could be hard. That could be left as future work, as the raw PCAP file already is useful.

soulgalore:
I'll be busy with preparing the coming sitespeed.io release the coming weeks, but if you have time @marty90 to do a PR of what it would look like, it would be interesting!

marty90:
I can try to implement it ;)
I may create a class `tcpdump` in `lib/support` used by the method `run` of the class `Engine`, like it does for `xvfb`. What do you think?

soulgalore:
@marty90 have a go if you have time and then we can help out in a PR to find a good structure. best to test out and if you find it useful we can help you get it in.

soulgalore:
This was released today in 8.0.0. `--tcpdump`

PR #604, opened 2018-05-24
Added PCAP trace capture feature

marty90:
As [previously](https://github.com/sitespeedio/browsertime/issues/591) discussed, I try a PR for the PCAP capture feature.

**How:**
* `Browsertime` starts a `tcpdump` child process that captures all traffic. So far, it captures traffic from **any** interface. It assumes that you machine/container is not doing other work, as its traffic may be captured.

**Dependencies:**
* TCPDump: the feature uses `tcpdump`. As `tcpdump` requires `root`, you need to set `SUID` bit to the `tcpdump` executable. This issue should not be present in the docker container.

**Command Line:**
* Added the `--pcap` option.

**Output:**
* This feature produces one `pcap` file per run. They are stored in the results dir under a subdirectory called `capture`. Each run creates a pcap file, named `<iteration>.pcap`.
E.g., with 3 runs I get:
```
capture/
├── 1.pcap
├── 2.pcap
└── 3.pcap
```

**Code:**
* I created a file in `lib/support` called `tcpdump.js` that is a wrapper for the `tcdump` executable
* It is called by `lib/core/engine/iteration.js` in the method `run`, to start/stop a capture at each run.

### DIFF
--- a/lib/core/engine/iteration.js
+++ b/lib/core/engine/iteration.js
@@ -9,6 +9,7 @@ const setResourceTimingBufferSize = require('../../support/setResourceTimingBuff
 const filterWhitelisted = require('../../support/userTiming').filterWhitelisted;
 const engineUtils = require('../../support/engineUtils');
 const Video = require('../../video/video');
+const TCPDUMP = require('../../support/tcpdump');
 const setOrangeBackground = require('../../video/screenRecording/setOrangeBackground');
 const stop = require('../../support/stop');
 const {
@@ -89,6 +90,11 @@ class Iteration {
 
     log.info('Testing url %s iteration %s', url, index);
     try {
+
+      // Start TCPDUMP
+      const myTCPDUMP = new TCPDUMP(this.storageManager.directory, this.options, index);
+      myTCPDUMP.start();
+
       await browser.start();
       await setResourceTimingBufferSize(browser.getDriver(), 600);
       await extensionServer.setupExtension(url, browser);
@@ -185,9 +191,18 @@ class Iteration {
       if (recordVideo) {
         await video.postProcessing(result);
       }
+
+      // Stop TCPDUMP
+      myTCPDUMP.stop();
+
       return result;
     } catch (e) {
       log.error(e);
+
+      // Stop TCPDUMP in case of error
+      if (myTCPDUMP)
+        myTCPDUMP.stop();
+
       // Take a screenshot in error case for verbose > 1
       if (options.verbose >= 1) {
         try {
--- a/lib/core/engine/iteration.js
+++ b/lib/core/engine/iteration.js
@@ -88,11 +88,16 @@ class Iteration {
       await video.setupDirs(index);
     }
 
+    // Instantiate a TCPDUMP instance
+    const myTCPDUMP = new TCPDUMP(
+      this.storageManager.directory,
+      this.options,
+      index
+    );
+
     log.info('Testing url %s iteration %s', url, index);
     try {
-
       // Start TCPDUMP
-      const myTCPDUMP = new TCPDUMP(this.storageManager.directory, this.options, index);
       myTCPDUMP.start();
 
       await browser.start();
@@ -200,8 +205,7 @@ class Iteration {
       log.error(e);
 
       // Stop TCPDUMP in case of error
-      if (myTCPDUMP)
-        myTCPDUMP.stop();
+      if (myTCPDUMP) myTCPDUMP.stop();
 
       // Take a screenshot in error case for verbose > 1
       if (options.verbose >= 1) {
--- a/lib/support/tcpdump.js
+++ b/lib/support/tcpdump.js
@@ -1,0 +1,40 @@
+'use strict';
+
+const { spawn } = require('child_process');
+const get = require('lodash.get');
+const mkdirp = require('mkdirp');
+
+/**
+ * Create a new TCPDUMP instance
+ * @class
+ */
+class TCPDUMP {
+  constructor(directory, options, iteration) {
+    this.directory = directory;
+    this.options = options;
+    this.iteration = iteration;
+  }
+
+  start() {
+    const useTcpDump = get(this.options, 'pcap', false);
+    if (useTcpDump === true || useTcpDump === 'true') {
+
+      // Create Capture dir
+      const capture_dir = this.directory + "/capture";
+      mkdirp(capture_dir);
+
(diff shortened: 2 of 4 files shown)


soulgalore:
Hey @marty90 
cool, `npm run lint:fix` should fix the most linting errors.

>  This issue should not be present in the docker container.

So does it work in Docker automatically? Sudo is present for the user in the container if I remember correctly (we have to be able to change network conditons).

Me or @tobli can do a review tomorow. There are a couple of things we can reuse, but I will show you in the review!

Thanks!
Peter

marty90:
`sudo` is not needed. `tcpdump` is executed without root.
I updated the `dockerfile` to install `tcpdump` so it should OK ;)

lib/core/engine/iteration.js
```diff
@@ -9,6 +9,7 @@ const setResourceTimingBufferSize = require('../../support/setResourceTimingBuff
 const filterWhitelisted = require('../../support/userTiming').filterWhitelisted;
 const engineUtils = require('../../support/engineUtils');
 const Video = require('../../video/video');
+const TCPDUMP = require('../../support/tcpdump');
```
soulgalore: Nitpick: We try to not overuse capital letters, so what about either TCPDump or TcpDump?

lib/core/engine/iteration.js
```diff
@@ -87,8 +88,18 @@ class Iteration {
...
+    const myTCPDUMP = new TCPDUMP(
+      this.storageManager.directory,
+      this.options,
+      index
+    );
+
     log.info('Testing url %s iteration %s', url, index);
     try {
+      // Start TCPDUMP
+      myTCPDUMP.start();
```
soulgalore: Can you add the check useTcpDump === true here? I think we most try to do the check first before using the function (could be that we have missed that sometimes), think it makes the code easier to read, you could follow the full flow in Iteration.

lib/core/engine/iteration.js
```diff
@@ -185,9 +196,17 @@ class Iteration {
       if (recordVideo) {
         await video.postProcessing(result);
       }
+
+      // Stop TCPDUMP
+      myTCPDUMP.stop();
```
soulgalore: Same here, only if used.

lib/support/tcpdump.js
```diff
@@ -0,0 +1,44 @@
+'use strict';
+
+const { spawn } = require('child_process');
```
soulgalore: We mostly try to use [execa](https://www.npmjs.com/package/execa) since it supports promises and have some better defaults than spawn.

lib/support/tcpdump.js
```diff
@@ -0,0 +1,44 @@
...
+const get = require('lodash.get');
+const mkdirp = require('mkdirp');
```
soulgalore: You can pass on the storage manager from Iteration and then create the dir with https://github.com/sitespeedio/browsertime/blob/master/lib/support/storageManager.js#L54 - then we keep making dirs in the StorageManager

lib/support/tcpdump.js
```diff
@@ -0,0 +1,44 @@
...
+  start() {
+    const useTcpDump = get(this.options, 'pcap', false);
+    if (useTcpDump === true || useTcpDump === 'true') {
+      // Create Capture dir
+      const capture_dir = this.directory + '/capture';
+      mkdirp(capture_dir);
+
+      // Start TcpDump
+      // Sould set suid bit with "sudo chmod +s /usr/sbin/tcpdump" if running outside docker
+      const capture_file = capture_dir + '/' + this.iteration + '.pcap';
```
soulgalore: You could probably use path join here to join the dirs (https://nodejs.org/api/path.html#path_path_join_paths) then at least in theory it will work on Windows. We don't support Windows but we try to keep the code working.

lib/support/tcpdump.js
```diff
@@ -0,0 +1,44 @@
...
+  constructor(directory, options, iteration) {
+    this.directory = directory;
+    this.options = options;
+    this.iteration = iteration;
+  }
+
+  start() {
+    const useTcpDump = get(this.options, 'pcap', false);
+    if (useTcpDump === true || useTcpDump === 'true') {
+      // Create Capture dir
```
soulgalore: Add a info log here and when we stop, it's perfectly ok I think, then it's so much easier to follow the flow what happens.

lib/support/tcpdump.js
```diff
@@ -0,0 +1,44 @@
...
+      const capture_dir = this.directory + '/capture';
```
soulgalore: Sorry we miss a style guide. capture_dir -> captureDir

lib/support/tcpdump.js
```diff
@@ -0,0 +1,44 @@
...
+      mkdirp(capture_dir);
+
+      // Start TcpDump
+      // Sould set suid bit with "sudo chmod +s /usr/sbin/tcpdump" if running outside docker
+      const capture_file = capture_dir + '/' + this.iteration + '.pcap';
+      this.tcpdump_process = spawn('tcpdump', [
```
soulgalore: tcpdump_process -> maybe tcpDumpProcess?

marty90:
Cool! Thank you! I'll fix the code soon ;)

soulgalore:
@marty90 had a short chat with @tobli he was more in the favor of adding this functionality as pre/post scrips. It should work I think, only thing we need to do is make the postscript execution a little more tighter (so it executes if you get some error). Doing it that way, we can let people try it out first and then if it is indemand, we incorporate it, but let @tobli describe how he thinks!

phongiswindy:
I would love to have this feature incooperated in browser time. Currently, I have to write separate code to monitor new docker interface and sniff from there, which sometimes very troublesome.

soulgalore:
Yes. I wouldn't mind integrating this without the pre/post script but just fixing the comments I made. Any chance you @phongiswindy could fix that? :)

phongiswindy:
> Yes. I wouldn't mind integrating this without the pre/post script but just fixing the comments I made. Any chance you @phongiswindy could fix that? :)

I really would like to help, but unfortunately I am not a javascript expert. I am afraid that I would mess things up rather then helping if I touch the code :)

soulgalore:
I'm closing this in the favor of #1159 (there are so many things that changed in the code base since then. You will get a special mention in the changelog @marty90  :)

marty90:
Great! Thank You!

Issue #610, opened 2018-06-14
Clear the console log before first access


soulgalore:
If you use the browsertime extension, for example doing a hard clear cache, that log messages ends up on the result page of the log. I've missed to clear the console log before we access the URL (as we do with the performance logs).

soulgalore:
fixed.

PR #611, opened 2018-06-14
Clear console log between runs #610

soulgalore:


lib/chrome/webdriver/chromeDelegate.js
```diff
@@ -49,6 +49,9 @@ class ChromeDelegate {
     // that can be fixed by emptying the logs twice :|
     await runner.getLogs(Type.PERFORMANCE);
     await runner.getLogs(Type.PERFORMANCE);
+    // Also make sure we clear the console log
+    // Hopefully one time is enough?
+    await runner.getLogs(Type.BROWSER);
```
tobli: 👍
This should be independent of skipHar, so maybe move it to the top of the function. I see there's similar issues with onStopIteration below.

soulgalore: Ah good catch!

soulgalore: yep let fix the other thing in another PR

Issue #622, opened 2018-07-20
Navigation fails


soulgalore:
On our test server we sometimes get errors that look like this:
`Failed to load/verify https://en.wikipedia.org/wiki/Barack_Obama uri:data:text/html;charset=utf-8,`

Other users too, see https://github.com/sitespeedio/sitespeed.io/issues/2097#issuecomment-406567113

This is because of the navigation didn't finish before we started to check, I thought I added a guard against that but maybe there's a bug there. 

ping @primehaxor

soulgalore:
Pushed a new version, hopefully this is fixed now.

PR #623, opened 2018-07-20
increase wait time for 5 to 50s (for really slow servers)

soulgalore:
I've seen reports in WPT where Chrome takes 40s to start to navigate so let us increase this limit
#622

### DIFF
--- a/lib/core/seleniumRunner.js
+++ b/lib/core/seleniumRunner.js
@@ -153,8 +153,8 @@ class SeleniumRunner {
         'return document.documentURI;'
       );
       await driver.executeScript(navigate);
-      // Wait max 5s on navigation
-      for (let i = 0; i < 10; i++) {
+      // Wait max 50s on navigation
+      for (let i = 0; i < 100; i++) {
         await delay(500);
         const newUri = await driver.executeScript(
           'return document.documentURI;'


skhilko:
@soulgalore, I've noticed that my scripts have slowed down quite a bit after this change. I assume, I'm missing a condition in either my login or log out scripts so that the full 50s timeout is executed. Could you point me to the direction where should I change my scripts so that the timeout can be broken ahead of time?

soulgalore:
Hi @skhilko ok, we still check every 500 ms if the navigation has happened? You can configure your own check with pageCompleteCheck but that will be for all of your runs. What we actually look for in the change is if the page URL has changed in the navigation (going from our internal start URL and then to the URL that you test). It works like this:
1. Ask the browser to navigate to the URL
2. Check if the URL has changed to URL
3. Loop to 2. until the URL has changed
4. Check if the page has finished loading using the pageCompleteCheck.
5. Loop to 4 until the pageCompleteCheck is done
6. Collect all the metrics.

Can you share more how you do and I can have a look? 
Best
Peter

skhilko:
@soulgalore, I think the problem is that the page we're measuring is automatically loaded after the login. What happens is:
1. Click Login button in the selenium script. I cannot verify that the user is logged in as suggested [here](https://www.sitespeed.io/documentation/sitespeed.io/prepostscript).
```
  // we wait for something on the page that verifies that we are logged in
  return driver.wait(until.elementLocated(By.id('pt-userpage')), 3000);
```
So I simply do
```
  return driver.findElement(By.css('#form-password [type="submit"]')).click();
```
2. The browser is automatically redirected to the page being measured (`LOGGED_IN_URL#dashboard`) after Login click.
3. The docker command happens to have the same test url `docker run ... --config config.json LOGGED_IN_URL/dashboard`

Am I doing this right?

soulgalore:
@skhilko hmm, ok. Let me update the documentation so it's easier to understand the flow.

If I understand your case correctly this is what happens:
1. You run your pre-script, where the user logs in and then gets redirected to X
2. Then the navigation happens in Browsertime, where browsertime asks the browser to navigate to X again. Since the check we have compare the URLs and in this case, they are the same it will take all the time until the timeout.

I think that the way it "worked" before wasn't either as you intended or? I mean the second access X (that is measured) I guess assets is cached (best case) in the browser cache, so you are measuring a user that first logs in and then hit the end of the redirect again? Was that understandable?

What we could do is change the check so it is a little bit smarter maybe ... let me look into that.

skhilko:
Not sure if I understand the last part of your message, but I think 1 and 2 are accurate.

In different words, I'm trying to measure the load time of the landing page after the log in. Is there a better way to do that?

soulgalore:
No I don't think so. I'll update the docs during the weekend so things are more clear.

PR #629, opened 2018-08-22
Add support for picking up visual element timings

soulgalore:
I've merged and cherry picked some changes in Visual Metrics to be able to collect timing metrics for individual elements.

You can try it out like this:
`docker run --shm-size=1g --rm -v "$(pwd)":/browsertime sitespeedio/browsertime https://en.wikipedia.org/wiki/Obama -n 1 --visualElements`

And if you want to pick your own elements (default get the largest image and largest h1):
`docker run --shm-size=1g --rm -v "$(pwd)":/browsertime sitespeedio/browsertime https://en.wikipedia.org/wiki/Obama -n 1 --visualElements --scriptInput.visualElements Logo:#p-logo`

I've added a new feature that maybe should have been in another PR:
You can now pass on parameters from the CLI to your own custom JS. --scriptInput.SCRIPT_NAME will match the SCRIPT_NAME to the filename of the script, and if it matches the params are passed as an args. We use it now to make it possible to pick individual elements on a page but I think this is something that really open up Browsertime to do even more great stuff.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,7 +62,7 @@ script:
 - node ./bin/browsertime.js -b chrome --skipHar -n 1 --preURL https://www.sitespeed.io/ -r header:value
   https://www.sitespeed.io/documentation/
 - node ./bin/browsertime.js -b chrome -vv --viewPort=640x480 --video --screenshot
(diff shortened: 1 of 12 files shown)


soulgalore:
At the moment we pick up only full visible images within the viewport but I think we should change that for mobile. The current solution means the Obama images isn't used:

soulgalore:
This is ready to go except that we need to tune picking up the largest image, seems to be something wrong there.

PR #660, opened 2018-11-01
Color change fix

soulgalore:


soulgalore:
Let us keep this open for a while. I'm testing getting the start point directly from VisualMetrics, it "works" but cutting the video goes wrong so need to finetune that.

soulgalore:
Fixed by https://github.com/sitespeedio/browsertime/pull/665

PR #664, opened 2018-11-04
New BT extension to support new blank URL used for mobile testing

soulgalore:


soulgalore:
Chrome on Android doesn't support extensions.

PR #673, opened 2018-11-13
Update Readme with the working code sample

Siwach16:
I am not sure if this is the right way (don't have any expertise in javascript) but thats how i made it work for my project

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,34 +185,9 @@ Testing a page after you have logged in:
 ~~~javascript
 module.exports = {
   run(context, help) {
-        await help.navigate(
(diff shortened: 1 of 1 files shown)


soulgalore:
Hi @Siwach16 thanks,
let me have a look early next week, I haven't looked at this for a while and want to finish it off next week.

Best
Peter

soulgalore:
I think there's some missunderstanding: There's different ways of measuring examples, either you can measure the login page (the actual login) or the page after you have logged in? In your example you change so it measure the logged in page instead of the next page. Let me update the docs so it's more clean.

soulgalore:
But you where right about not fully working examples, I updated them now but leaved out the change of functionality, thank you @Siwach16 for the PR!

PR #683, opened 2018-12-11
Specifically check if type of FF preference is a boolean value along with isNaN()

### LINKED ISSUE #684: boolean Firefox Preferences are not being set properly
Using --firefox.preference for boolean prefs such as "javascript.options.ion" does not currently work.  I hit this problem with and without docker on Linux Mint 19 using FF 63, 64 and 65.   I was just using reddit as an example URL, but any will reproduce the problem.

dpalmeiro:
Resolves #684 

boolean firefox preferences are not being set properly because they fail the isNaN() check below and therefore the preferences are set to 0 or 1 instead, which doesn't work.


soulgalore:
Hi @dpalmeiro thanks for the fix! Release coming up ... hopefully next week.

PR #702, opened 2019-01-02
Manually set properties httpProxyPort and sslProxyPort for Firefox

### LINKED ISSUE #701: Firefox proxy port not being set properly
When using the --proxy.http and --proxy.https options, the hostnames are set properly, ,but the ports are always set to 80, and 443 for http and SSL proxies.

A simple example:
browsertime https://www.sitespeedio.com --skipHar -b firefox --proxy.http localhost:8080 --proxy.https localhost:8080
If you navigate to preferences->network settings, you'll see that the ports are not set to 8080.  

This appears to be an existing bug in selenium:  https://github.com/SeleniumHQ/selenium/issues/5627

dpalmeiro:
Closes #701.  This provides a workaround that allows the proxy port to be set properly for Firefox until the problem in Selenium is addressed.

### DIFF
--- a/lib/firefox/webdriver/index.js
+++ b/lib/firefox/webdriver/index.js
@@ -132,7 +132,10 @@ module.exports.configureBuilder = function(builder, baseDir, options) {
   const proxySettings = pick(options.proxy, ['http', 'https']);
 
   if (!isEmpty(proxySettings)) {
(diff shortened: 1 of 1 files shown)


soulgalore:
Hi @dpalmeiro ah thanks for the fix!

PR #746, opened 2019-01-31
Click on whatever element (hidden or not)

soulgalore:


soulgalore:
I think it will be cleaner for now to just run JavaScript and wait for page load.

PR #762, opened 2019-02-07
Use alias instead of URL where possible

soulgalore:


soulgalore:
This way to complicated and need to come in next major.

PR #770, opened 2019-02-12
Print Correct Timestamp to Video

ponyo877:
The display of the video's Timestamp is  .00 -> .${framerate} -1 .
ex) When framerate is 30 (default), Timestamp is 0.00 -> 0.29, 1.00 -> 1.29, 2.00 to 2.29, ...
This Change is 0.00 -> 0.99, 1.00 -> 1.99, ...

### DIFF
--- a/lib/video/postprocessing/finetune/addTextToVideo.js
+++ b/lib/video/postprocessing/finetune/addTextToVideo.js
@@ -25,9 +25,7 @@ module.exports = async function(
         : '';
   args.push(
     '-vf',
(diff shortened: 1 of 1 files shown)


soulgalore:
Ah thanks for the fix @ponyo877 !

soulgalore:
@ponyo877 this was great! one small thing, the timing texts moves a little (left/right) small small move, any chance you can do follow up PR to fix it? 

Thanks
Peter

PR #773, opened 2019-02-14
Bugfix: New timestamp in the video introduced some flickering in the …

ponyo877:
…time stamp

If the font size is too large, the characters move.
When # 771 - borderw + (box, boxcolor), Desktop and mobile have stopped moving.


soulgalore:
Thanks @ponyo877 lets try this out, I did a fix yesterday that worked fine on my local my machine but pushing it to the test servers it started to jump again.

PR #789, opened 2019-02-22
use alias as foldername

vitaliytv:
When i use script with alias (not as second params):
```
module.exports = async function (context, commands) {
  await commands.measure.start('login')
  await commands.measure.start('https://www.sitespeed.io/examples/')
}
```

i got error:
```
[2019-02-22 22:17:09] ERROR: Error running browsertime TypeError: Cannot read property 'replace' of null
    at /usr/local/lib/node_modules/browsertime/lib/support/pathToFolder.js:42:37
    at Array.forEach (<anonymous>)
    at module.exports (/usr/local/lib/node_modules/browsertime/lib/support/pathToFolder.js:40:18)
    at Collector.perIteration (/usr/local/lib/node_modules/browsertime/lib/core/engine/collector.js:185:19)
    at <anonymous>
```
 this path  create result directory  with name = alias (without hostname)

Maybe create result folder for every  commands.measure.start call?

### DIFF
--- a/lib/core/engine/collector.js
+++ b/lib/core/engine/collector.js
@@ -92,7 +92,7 @@ class Collector {
   async perIteration(allData, index) {
     for (let data of allData) {
       const alias = data.alias;
-      let url = data.url || this.urlFromCli;
+      let url = data.url || alias || this.urlFromCli;
       if (alias && !this.aliasAndUrl[alias]) {
         this.aliasAndUrl[alias] = url;
       } else if (alias && this.aliasAndUrl[alias]) {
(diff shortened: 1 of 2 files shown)


soulgalore:
Hi @vitaliytv so I'm tired after a long a week and I'm not following it 100%, I need to look into if it will break things, or are you saying it's broken already? :)

Best
Peter

soulgalore:
I think so much has changed since this PR so closing it. In the long term we could change how create folders with names but we need to have a new long term strategy for that.

PR #834, opened 2019-05-11
Test out using CDP directly instead of Chromedriver to get the trace …

soulgalore:
…log.


soulgalore:
If this works out, we should do a follow up, cleanup.

Issue #755, opened 2019-02-04
Visual Metrics picks up small orange using Chrome


soulgalore:
Updated to the latest Visual Metrics we removed our fix for checking that there are no partial orange in the screen. 

![ms_000000](https://user-images.githubusercontent.com/540757/52197909-7ae0c380-2861-11e9-94b1-d0bbaea7bdc7.jpg)

This give some page a really early first visual change.

soulgalore:
Some fixes in https://github.com/sitespeedio/browsertime/pull/757 rolled out in 4.2.2.

soulgalore:
This is dependent on FPS. If we use 60 fps this happens often. 30 fps its better. And I guess 10 fps will eliminate it.

soulgalore:
Haven't seen this since the switch to 30 fps.

soulgalore:
I still see this on our test instance:
<img width="804" alt="Screen Shot 2019-06-08 at 19 59 38" src="https://user-images.githubusercontent.com/540757/59150703-02b12700-8a28-11e9-8572-71a7fec658d8.png">

soulgalore:
The 33 ms is a sign of that the small orange bit picks up the orange. I think we should fine tune Visual Metrics again.

soulgalore:
This is better now but still happens. We should do another fix in Visual Metrics so we the bottom area is larger and maybe fine tune the area that could be orange.

soulgalore:
This is still an issue:
![ms_000000](https://user-images.githubusercontent.com/540757/72890527-5bf96c80-3d12-11ea-840f-7454962261f3.jpg)

Let me try to reproduce with an original video so its easier to fix. Ping @tarekziade is this something you have seen? Maybe there's a smarter way than the fix I've been trying out.

tarekziade:
@sefeng211 do you have an idea?

sefeng211:
Sorry I was on PTO last week, I'll take a look sometime this week.

soulgalore:
Let me know @sefeng211 if you have time to take a look. I think somewhere there's bug either how we check "the bottom" of the screen or if it could be the margin that also messes it up. or if there's just a better way to check.

sefeng211:
@soulgalore Sorry, didn't get a chance to look it last week. And I misunderstood the bug, I thought it was related to the converting 60fps PR we merged. 
I don't think I fully understand what is happening. Is this Chrome desktop? And there are partial orange frame left when cropping? 
[...]

PR #854, opened 2019-06-10
Include last 50 pixel when checking for orange.

soulgalore:
When checking for orange frame, Visual Metrics don't include the bottom
50 px. That is probably ok if your video has the browser window at the
bottom but we crop that out.

We need to verify that this works on mobile too.


soulgalore:
Hopefully fixes https://github.com/sitespeedio/browsertime/issues/755

PR #868, opened 2019-06-28
One listener is enough

soulgalore:
See https://github.com/sitespeedio/sitespeed.io/issues/2513


soulgalore:
Lets come back later to this with a better fix.

PR #905, opened 2019-08-01
[6.0 and Chrome 77] Use Chrome layout shift (Chrome 77 will support buffered observer)

soulgalore:


soulgalore:
We can start test this when Chromedriver 77 is released.

PR #907, opened 2019-08-01
[6.0 and Chrome 77] Always get long task events (for Chrome).

soulgalore:
With Chrome 77 you can get buffered data from PerformanceObservers,
meaning we don't need to inject JS early when loading the page.


soulgalore:
If I understand correctly this isn't supported in Chrome 77 or even in the spec yet, so we have different behaviours for PerformanceObservers :( https://github.com/w3c/longtasks/issues/41

PR #976, opened 2019-10-28
Calculate contentful speed index using edge detection integration

dpalmeiro:
Contentful speed index is a new SI metric developed by Bas Schouten at Mozilla which uses edge detection to calculate the amount of "content" that is visible on each frame.   

It was primarily designed for two main purposes:
1.  Have a good metric to measure the amount of text that is visible.
2.  Design a metric that is not easily fooled by the pop up splash/login screens that commonly occur at the end of a page load.  These can often disturb the speed index numbers since the last frame that is being used as reference is not accurate.

### DIFF
--- a/lib/core/engine/collector.js
+++ b/lib/core/engine/collector.js
@@ -176,7 +176,7 @@ class Collector {
       if (data.visualMetrics) {
         if (this.options.iterations > 1) {
           log.info(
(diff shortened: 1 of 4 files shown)


soulgalore:
Thank you @dpalmeiro !

PR #978, opened 2019-10-28
Use the Firefox internal window recorder to calcualte visual metrics

dpalmeiro:
Firefox 67 and above has a built-in window recorder ([bug 1536174](https://bugzilla.mozilla.org/show_bug.cgi?id=1536174)) that is able to dump PNG images of each frame that is painted to the window.   This can be enabled and disabled in the browser console, or through the chrome context with selenium webdriver.

This PR introduces a new privileged API that is able to execute JS in the chrome context, as well as support for generating a variable rate MP4 using the output images from the window recorder.  

The motivation for this work was to introduce a low-overhead video recorder that will not introduce performance disturbances during page loads.

### DIFF
--- a/lib/core/engine/command/javaScript.js
+++ b/lib/core/engine/command/javaScript.js
@@ -49,7 +49,10 @@ class JavaScript {
    */
   async runPrivileged(js) {
     try {
-      const value = await this.browser.runPrivilegedScript(js, 'CUSTOM PRIVILEGED');
+      const value = await this.browser.runPrivilegedScript(
+        js,
+        'CUSTOM PRIVILEGED'
+      );
       return value;
     } catch (e) {
       log.error('Could not run privileged JavaScript %s ', js);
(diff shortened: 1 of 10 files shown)


soulgalore:
Think this needs an updated npm-shrinkwrap.json to pass in Travis.

dpalmeiro:
Thanks, I updated the version of del in package.json to sync up with the one in shrinkwrap.

soulgalore:
Thank you @dpalmeiro this will be great and I really love to try it out. For the Travis test to work, you could just add a extra check that the options have options.firefox, I think we need to fix our test but I can do that later.

Also if you could add one line in the .travis.yml file to test the new functionality that would be great! Something simple like:
`- node ./bin/browsertime.js -b firefox -n 1 --firefox.windowRecorder https://www.sitespeed.io`

I had a quick go at my local machine but couldn't get it to work, I got an error when trying to add the text to the video, but that could be because my local is missing some dependency, will have a look when I'm back from vacation on Friday.

soulgalore:
Yeah that bug is on my side. I have something wrong with my ffmpeg version, so that running with VisualMetrics fails. Using just the window recorder with video works fine for me.

soulgalore:
I'll take this home.

soulgalore:
Thank you @dpalmeiro :)

PR #995, opened 2019-11-19
Use async composition recorder API from Firefox 72

brennie:
Firefox Nightly 72 now uses an asynchronous
`windowUtils.setCompositionRecording` API. This unfortunately breaks
Browsertime on Firefox 72+.

However, we can support both Firefox <= 71 and >= 72 by wrapping the
return value of `windowUtils.setCompositionRecording(enable)` in
`Promise.resolve()`. When `Promise.resolve()` is passed a promise, the
returned promise will resolve immediately, which solves the problem for
the synchronous API in Firefox <= 71. However, when passed a promise,
`Promise.resolve()` will return the passed promise, which allows us to
support both APIs.

### DIFF
--- a/lib/video/screenRecording/firefox/firefoxWindowRecorder.js
+++ b/lib/video/screenRecording/firefox/firefoxWindowRecorder.js
@@ -9,18 +9,29 @@ const del = require('del');
 
 const unlink = util.promisify(fs.unlink);
 
-async function enableWindowRecorder(enable, driver) {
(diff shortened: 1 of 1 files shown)


soulgalore:
Ah cool @brennie ! So we don't auto-lint in Travis (sorry), can you just fix it and re-push?

Best
Peter

soulgalore:
Thank you @brennie !

PR #999, opened 2019-11-27
try out running more tests headless

soulgalore:


soulgalore:
This has changed so much so lets just skip it.

Issue #1027, opened 2019-12-12
Window recorder doesn't work


soulgalore:
Testing on Firefox 72:
` --firefox.windowRecorder --video https://www.sitespeed.io -n 1 -b firefox`

generates 

```
[2019-12-12 07:57:40] ERROR: JavascriptError: ReferenceError: windowUtils is not defined
    at Object.throwDecodedError (/Users/phedenskog/git/browsertime/node_modules/selenium-webdriver/lib/error.js:550:15)
    at parseHttpResponse (/Users/phedenskog/git/browsertime/node_modules/selenium-webdriver/lib/http.js:560:13)
    at Executor.execute (/Users/phedenskog/git/browsertime/node_modules/selenium-webdriver/lib/http.js:486:26)
    at process._tickCallback (internal/process/next_tick.js:68:7)

```

Hi @brennie I think something went wrong with https://github.com/sitespeedio/browsertime/pull/995 ?

brennie:
Hi! I'm investigating this and will hopefully have a fix to you today.

soulgalore:
Fixed in https://github.com/sitespeedio/browsertime/pull/1037

Issue #1013, opened 2019-12-04
Contentful Speed Index generates errors


soulgalore:
Hi @dpalmeiro could you help me with this? I've got some reports on Slack of errors from Visual Metrics that I haven't been able to reproduce until now. I got a video file and the log. The error:

```
Traceback (most recent call last):
  File "./visualmetrics.py", line 1875, in main
    options.herodata)
  File "./visualmetrics.py", line 1324, in calculate_visual_metrics
    'value': calculate_contentful_speed_index(progress, dirs)})
  File "./visualmetrics.py", line 1477, in calculate_contentful_speed_index
    value  = int(output.split()[7].split(':')[0])
IndexError: list index out of range
```
I can reproduce it like this: 
`$browsertime/vendor ./visualmetrics.py -i error.mp4  --orange --perceptual --contentful --force --renderignore 5 --json --viewport -vvv`

Removing the Contentful switch makes it work again.

https://send.firefox.com/download/705cdf51d7b31083/#SFzPnGSrl8GqP_Hi1APP6g

Best
Peter

soulgalore:
I wrapped the code in a try/catch now and then we can take on the real issue in the future. This makes the rest of the Visual Metrics work.

soulgalore:
This is fixed in 8.0.0

Issue #983, opened 2019-11-04
window recorder for Firefox gives HAR errors


soulgalore:
Since I enabled the window recorder on our test server, I see many of these kind of errors:

```
[2019-11-04 06:18:48] INFO: Running tests using Firefox - 3 iteration(s)
[2019-11-04 06:18:51] INFO: Testing url https://en.wikipedia.org/wiki/Main_Page iteration 1
[2019-11-04 06:18:51] INFO: Start firefox window recorder.
[2019-11-04 06:18:58] INFO: Stop firefox window recorder.
[2019-11-04 06:18:59] INFO: Waiting on har-export-trigger to collect the HAR
[2019-11-04 06:18:59] ERROR: Could not get the HAR from Firefox JavascriptError: ReferenceError: HAR is not defined
    at Object.throwDecodedError (/usr/src/app/node_modules/selenium-webdriver/lib/error.js:550:15)
    at parseHttpResponse (/usr/src/app/node_modules/selenium-webdriver/lib/http.js:560:13)
    at Executor.execute (/usr/src/app/node_modules/selenium-webdriver/lib/http.js:486:26)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at async Driver.execute (/usr/src/app/node_modules/selenium-webdriver/lib/webdriver.js:699:17)
    at async FirefoxDelegate.onCollect (/usr/src/app/node_modules/browsertime/lib/firefox/webdriver/firefoxDelegate.js:80:25)
    at async Measure.collect (/usr/src/app/node_modules/browsertime/lib/core/engine/command/measure.js:296:5)
    at async /usr/src/app/node_modules/browsertime/lib/core/engine/run.js:9:9
    at async Iteration.run (/usr/src/app/node_modules/browsertime/lib/core/engine/iteration.js:162:9)
    at async Engine.runByScript (/usr/src/app/node_modules/browsertime/lib/core/engine/index.js:121:20)
    at async Object.analyzeUrl (/usr/src/app/lib/plugins/browsertime/analyzer.js:125:21)
    at async Object.processMessage (/usr/src/app/lib/plugins/browsertime/index.js:134:24)
```

soulgalore:
I think this is memory related. When I switched to run fewer tests, it seems that we don't get them, so often (or at all).

soulgalore:
Also this:
```
[2019-11-04 13:55:11] INFO: Start firefox window recorder.
[...]

PR #1028, opened 2019-12-12
Merge mozilla-browsertime changes into sitespeedio-browsertime

gmierz:
This PR is adding the commits from mozilla-browsertime into the sitespeedio-browsertime repo (upstreaming the changes).

The following files had merge conflicts so they are of most interest in the reviews:
```
U	lib/android/index.js
U	lib/chrome/webdriver/index.js
U	lib/core/defaultPageCompleteCheck.js
U	lib/core/engine/collector.js
U	lib/core/engine/command/javaScript.js
U	lib/core/engine/command/measure.js
U	lib/core/engine/index.js
U	lib/core/engine/iteration.js
U	lib/core/seleniumRunner.js
U	lib/firefox/webdriver/firefoxDelegate.js
U	lib/firefox/webdriver/firefoxPreferences.js
U	lib/firefox/webdriver/index.js
U	lib/screenshot/index.js
U	lib/support/cli.js
U	lib/support/har/index.js
U	lib/video/screenRecording/firefox/firefoxWindowRecorder.js
U	lib/video/screenRecording/recorder.js
U	lib/video/video.js
U	npm-shrinkwrap.json
U	package.json
U	vendor/visualmetrics.py
```


gmierz:
TODO: Remove this commit from the upstreamed changes: https://github.com/sitespeedio/browsertime/pull/1028/commits/ec43310dedc2e9b2e49574edb5043f483dcf34d3

soulgalore:
Wow this will be awesome! I normally don't merge so large PRs though, but understand that's almost the only way so lets go! I think this will need to go as a new major release, but thats no problem.

I need to setup a server to continuously run some tests (as we do with sitespeed.io) before we merge, I'll do that tomorrow. Wanna have it running some time before we merge so we easily can find things (there are couple of issues from the latest merges #1027 #1013 #983 , so it could be that we find more things when we merge). 

Let me know when you think is ready for me to look into the PR!

soulgalore:
I've setup an instance but it seems the autobuild on Docker doesn't include your PR; maybe only branches in the project. My idea was to just switch to an auto build there, but I guess we could just build this branch on the server.

Can you @gmierz make it go through the Travis tests `npm run lint:fix` + there are some things breaking, then I can make it run there + go through the commits, I think I have a couple of questions :)

gmierz:
Sounds good - I'm happy to find a way to split this up if needed.

@soulgalore, the geckoprofiler changes have the `fs` package as a dependency but I see that it is a restricted module. Would you know of an alternative we could use here? It's being used here right now: https://github.com/sitespeedio/browsertime/blob/5f97b5c832bca0933a204372d58a54c1211f52c1/lib/core/engine/iteration.js#L237

soulgalore:
No lets go ahead with this @gmierz thats ok :)

My plan is to review tomorrow and then hopefully merge the coming week so that is fixed before Christmas.

> is a restricted module.

No you could use that or?

gmierz:
Ok, I ask because it's one of the linting failures: https://travis-ci.org/sitespeedio/browsertime/builds/624336797#L768

I would have to remove `fs` from here to use it - if that's ok with you then I'll make the change: https://github.com/sitespeedio/browsertime/blob/1ef54896a708949ed3d858784de71a3137b6f626/lib/core/.eslintrc.json#L3

soulgalore:
> I would have to remove fs from here to use it - if that's ok with you then I'll make the change:

Ah I see. I think the reason is that we used that to make sure we moved all file handling to the storage manager (my guess is that from the beginning it was all over the place). Either we can just add a utility method or as you said just remove that rule, do what you think is best :)

gmierz:
@soulgalore a small update on progress.
1. Added a method to StorageManager to be able to read files (for the geckoprofiler).
2. Added a fix for the geckodriver not being found by the selenium-webdriver.
3. Added a fix to check if an extra field exists before it is used (we added those changes).

At the moment I'm looking into an issue where the test scripts are being recognized as problematic and are hitting this error: https://github.com/sitespeedio/browsertime/pull/1028/files#diff-598e2038200e98260b612f9a4fc397d7R560

I think I will have to change the test scripts being used but maybe there's another way.

gmierz:
I found a better solution for that last problem. :)

soulgalore:
I looked into the problem and the issue is that we do multiple calls to video.stop. That part was moved on the main branch to happen before we collect all video metric I guess that caused the merge to add it twice. I didn't have push rights to your repo so this is the change that is needed:

Replace https://github.com/gmierz/browsertime/blob/mozbtimesync/lib/core/engine/command/measure.js#L270-L273

with:
https://github.com/gmierz/browsertime/blob/mozbtimesync/lib/core/engine/command/measure.js#L315-L325

lib/core/defaultPageCompleteCheck.js
```diff
@@ -2,9 +2,9 @@
 
 module.exports = `
 return (function(waitTime) {
-    try { 
-        const end = window.performance.timing.loadEventEnd;
-        return (end > 0) && (Date.now() > end + waitTime);
+    try {
```
soulgalore: So we skip the extra wait time now, is there a reason for that? That is simple wait for the user to add more wait instead of changing the more complicated pageCompleteCheck

lib/core/defaultPageCompleteCheck.js
```diff
@@ -2,9 +2,9 @@
...
+        var end = window.performance.timing.loadEventEnd;
+        return end > 0;
```
soulgalore: We skip the use of the extra waitTime, any reason for that? That's an way for the user to make the test test longer without using the more complicated pageCompleteCheck.

lib/core/seleniumRunner.js
```diff
@@ -425,9 +427,18 @@ class SeleniumRunner {
         await this.driver.quit();
         log.debug('Closed the browser.');
       } catch (e) {
-        throw new BrowserError(e.message, {
-          cause: e
-        });
+        // Work around GeckoView Bug 1298921: https://bugzilla.mozilla.org/show_bug.cgi?id=1298921.
+        const bug_1298921 =
```
soulgalore: Haha I love practical programming :D

lib/firefox/webdriver/firefoxPreferences.js
```diff
@@ -12,35 +12,9 @@ module.exports = {
   // Use hidden TTCP
   'dom.performance.time_to_contentful_paint.enabled': true,
```
soulgalore: What about when we do this release switch to use your defaults instead and release it as a major release? Right now these are` --firefox.mozillaProPreferences` but I think we can just make those default in later PR or do you see a problem with that?

vendor/visualmetrics.py
```diff
@@ -1521,11 +1544,18 @@ def calculate_perceptual_speed_index(progress, directory):
...
+    raw_progress_value = ["0=0"]
+    for timestamp, percent in completeness_value:
+        p = int(100 * float(percent) / float(per_si))
+        raw_progress_value.append('%d=%d' % (timestamp, p))
+
+    return per_si, ", ".join(raw_progress_value)
 
 
 def calculate_hero_time(progress, directory, hero, viewport):
-  try:
```
soulgalore: The try/except needs to be included in the PR. We wrap the hero times because it fails on Arabic Wikipedia making all visual metrics fail. Think try/except is needed both hereto and Contentful speed index

lib/firefox/webdriver/firefoxPreferences.js
gmierz: I'll undo these changes and we can do that in a later PR.

lib/core/defaultPageCompleteCheck.js
gmierz: The reason we removed it here is because when we playback a recorded webpage (through mitmproxy) on Chrome, this code hangs. We moved this wait to here - I can change it back or we can try to find another solution to this: https://github.com/sitespeedio/browsertime/pull/1028/files/88bd5c79e52e2fb4b1ec2263bb1510666fd41ca2#diff-598e2038200e98260b612f9a4fc397d7R178

soulgalore: Ah I see, we do the same for WebPageReplay since the date object is static: https://github.com/sitespeedio/browsertime/blob/master/docker/scripts/start.sh#L91 

Push this as is and the I can change the code so we use performance.now instead (or do you also override that?). Since we plan to release this as a major, it would be nice to fix so it works out of the box both and without a reply proxy, that will make things simpler.

gmierz:
Thanks @soulgalore for pointing me to the issue there! I'll update the patch with those changes.

I'm currently wrangling with some issues running this merge in production. For some reason the preferences we provide in a browser profile are being ignored and because of this we don't get many metrics (this test run only has the "loadtime" metric): https://treeherder.mozilla.org/#/jobs?repo=try&tier=1%2C2%2C3&revision=7cc129553a3b180855dba070a69291a31095982b&selectedJob=281196445

The reason I say that the preferences are being ignored is because when I add the preferences we need into the default Firefox Browsertime preferences, we get the metrics even though they've already been provided within the profile.

lib/core/defaultPageCompleteCheck.js
gmierz: Sounds good, and I don't think we override performance.now anywhere.

soulgalore:
@gmierz I think you can move https://github.com/sitespeedio/browsertime/blob/master/lib/firefox/webdriver/firefoxPreferencesMozillaPerf.js them to the default settings in your PR just to make it easy, then it "should" just work.

vendor/visualmetrics.py
gmierz: Ok, I've added this back. The contentful speed index function has it as well.

gmierz:
@soulgalore I ended up using the new `--firefox.mozillaProPreferences` for that and it's working now. :)

Here's the latest test run: https://treeherder.mozilla.org/#/jobs?repo=try&tier=1%2C2%2C3&revision=f5cd6290da0161746dac6a6d8561df664c31d633&selectedJob=281197852

soulgalore:
Lets fix what's missing in master :)

PR #1048, opened 2019-12-21
Add await in runPrivilegedScript to fix race condition.

gmierz:
This patch fixes an issue with gecko profiling where a race condition occurs. The issue happens because the context gets changed from chrome to content before the script has had a chance to run (which requires the chrome context).

It fixes this error: 
```
[2019-12-21 00:43:40] INFO: Testing url https://www.sitespeed.io iteration 1
[2019-12-21 00:43:40] ERROR: No page that could be associated with the error:ReferenceError: Services is not defined
[2019-12-21 00:43:40] ERROR: JavascriptError: ReferenceError: Services is not defined
```

### DIFF
--- a/lib/core/seleniumRunner.js
+++ b/lib/core/seleniumRunner.js
@@ -319,7 +319,7 @@ class SeleniumRunner {
       try {
         await this.driver.setContext('chrome');
 
-        return timeout(
+        return await timeout(
           this.driver.executeScript(script, args),
           scriptTimeout,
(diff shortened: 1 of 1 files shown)


soulgalore:
Ah sorry @gmierz I didn't check the PRs and looked into it just now in https://github.com/sitespeedio/browsertime/pull/1049 - I'll do better next time. 

I'll keep my fix since also simplified the code. Let me now if you disagree!

gmierz:
No problem :) I'll close this now since one of your commits fixed the issue.

PR #1050, opened 2019-12-22
Start Geckoprofile per URL instead of for the first iteration.

soulgalore:


soulgalore:
Hi @gmierz the old implementation started Geckoprofile on the first iteration but I think its better to start per URL instead?

There's a couple of things here: The names of methods of the delegates aren't clear, they have evolved and its not easy to understand when they run. Lets rename them before the next release

If you use scripting, you could test multiple URLs in the same iteration and with the old approach we start the profile at the beginning and stop it after each URL (onCollect) that's why I moved it to start the profile per URL we test instead.

Does this looks good for you? I'm not gonna merge until you have had a look.

gmierz:
Hi @soulgalore, these changes work with both warm and cold pageload testing, which is great!

One issue I found is that the profile folders are organized differently between warm and cold.

On warm runs, they are as expected (or the same as they were before these changes) but when running cold tests, the profiles get added to the same query folder, and there's also an extra profile that is being produced now.

That said, this issue is unrelated to this PR and has been around for a while (it occurs on our 4.9.2 fork as well) - there's no need to fix it here. The patch in this PR looks great to me. :) 

Profiles for warm page load:
image

Profiles for cold pageload:
image

soulgalore:
Aha ok. The extra ones on warm is because you use the same URL inside a script file. We maybe coyld fix that, but have you looked into if you could use —preURL for warming? I think that should fix it.

soulgalore:
The reason that it gets different paths is the use case where you for example login the user and then hit the same page twice, then we wants to seperate it. You could try —preURL with the same URL as the URL you want yo test. Then the browser it kept open between the preURL and the URL that gets tested, and the browser cache is kept between them @gmierz

gmierz:
Oh cool, thanks! I'll give that shot - it shouldn't prevent this patch from landing though.

PR #1053, opened 2019-12-25
Remove await async/pattern

soulgalore:
See ##1052


soulgalore:
We could simplify the code more but let us do that in follow ups.

Issue #1082, opened 2020-01-08
Enable verbose logging when using geckodriver


ncalexan:
`selenium-webdriver` 4.0.0-alpha1 added an equivalent of setting the default service builder, namely `setFirefoxService`.  This finally allows browsertime to configure geckodriver logging (and to set the path to geckodriver itself directly).

soulgalore:
Merged, released in next alpha.

PR #1083, opened 2020-01-08
Enable verbose logging when using geckodriver. (#1082)

ncalexan:
This also configure the path to the geckodriver binary without
manipulating PATH.


soulgalore:
Hi @ncalexan thanks, that's great!

Issue #1081, opened 2020-01-08
Make binary dependencies `optionalDependencies`


ncalexan:
Hi @soulgalore!  nalexander from Mozilla here.  To install browsertime in Mozilla's automation we package a `node_modules` directory into an archive and unpack it, including across different machine architectures.  That works as long as the `node_modules` doesn't contain binary/compiled dependencies (which are arch-specific).  To work around this in https://github.com/mozilla/browsertime, I made the binary dependencies `optionalDependencies` and worked around them being missing at runtime:
- https://github.com/mozilla/browsertime/commit/7d9db44ac746d18a93c0ca7d9e2600b49466593c
- https://github.com/mozilla/browsertime/commit/0e9ed0a4c65910c266a45db5f303274f0ce1db0d

Then in automation we don't install optional dependencies.

This solution works well for us but imposes a slight penalty on the repo.  I think it doesn't impose a penalty on users since `npm install` installs optional dependencies by default.

How do you feel about this?  Have you other thoughts on how to achieve this?  The only alternative I can think of is to manually splice a fake source-only package into the `node_modules` directory in place of the binary dependencies in our automation archives, which isn't great.  (Or build separate archives per architecture, which isn't feasible for us due to macOS.)

soulgalore:
Hi @ncalexan for Geckodriver and Chromedriver we have a way of skipping installation by setting a variable on installation see https://www.sitespeed.io/documentation/sitespeed.io/installation/#skip-installing-chromedrivergeckodriver - that way you could run npm install but skipping installing the drivers. 

But moving the three of them to optional and use `--no-optional` would be great I think. I think we should add some logging  in the catch block for sharp, just a info log that we run without sharp. I wanna test also if we can do something smart if we run without sharp so most work.

soulgalore:
Hi @ncalexan so I merged moving Sharp as an optional dependency and make it work without it (but then just producing PNG files). 

Would using the flag/variables be enough for you to skip installing Gecko/Chromedriver? The reason is that moving them to optional would make the npm install signal an ok installation even though one of them fail to install.

ncalexan:
> Hi @ncalexan so I merged moving Sharp as an optional dependency and make it work without it (but then just producing PNG files).
> 
> Would using the flag/variables be enough for you to skip installing Gecko/Chromedriver? The reason is that moving them to optional would make the npm install signal an ok installation even though one of them fail to install.

[...]

PR #1084, opened 2020-01-09
Make Sharp an optional req #1081

soulgalore:
We can tune this (move out sharp to only require it once) but I think this could be good enough as a start. With this fix screenshots works without sharp but they are saved only as png and at viewport size.


soulgalore:
Lets move this on so we can continue with Gecko/Chrome-drivers.

PR #1089, opened 2020-01-10
Fix font usage on macOS

tarekziade:
On macOS, addTextToVideo and getTimingMetrics will fail because the code
makes the assumption /System/Library/Fonts/SFNSText.ttf always exists.

In the new getFont() api, we check that assumption and fallback to
not specifying a font. Also, we make sure to use the font passed
through options.videoParams.fontPath even on macOS because why not.


tarekziade:
I should mentioned that I don't have that font on Catalina -- so browsertime was failing for me there

soulgalore:
Ah cool, thanks @tarekziade ! I'll merge this asap when Travis passes.

soulgalore:
Thanks again @tarekziade !

Issue #1109, opened 2020-01-13
Teach browsertime about brand new `--firefox.env` option for setting environment variables


ncalexan:
Cutting edge (re, trunk) `geckodriver` builds will support configuring the runtime environment, including on Android, now that I've implemented https://bugzilla.mozilla.org/show_bug.cgi?id=1607960.  This ticket tracks exposing `--firefox.env` as part of browsertime, just like we can set `--firefox.prefs`, etc.

There's not (yet) support for this in `selenium-webdriver`, so this will be another "side-around".

soulgalore:
This has been fixed.

PR #1110, opened 2020-01-14
Add `--firefox.args` and `--firefox.env` options for arguments and en…

ncalexan:
…vironment variables. (#1109)

`--firefox.args` is sibling to `chrome.args`; support for this already
exists in `selenium-webdriver`.

`--firefox.env` has no analog, and requires a `geckodriver` with
https://bugzilla.mozilla.org/show_bug.cgi?id=1607960 applied.  That
will likely be whatever release follows 0.26.0.

### DIFF
--- a/lib/firefox/webdriver/index.js
+++ b/lib/firefox/webdriver/index.js
@@ -150,6 +150,21 @@ module.exports.configureBuilder = function(builder, baseDir, options) {
 
   ffOptions.addArguments('-no-remote');
 
+  if (firefoxConfig.args) {
+    ffOptions.addArguments(firefoxConfig.args);
+  }
+
+  const envMap = {}; // Map of string name to string value.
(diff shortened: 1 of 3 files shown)


soulgalore:
Looks cool, what about hiding just the cli for `--firefox.env` until its usable = next Geckodriver or will it be released soon do you know? I wanna get out the current things in master asap, but I guess that probably would be next week.

ncalexan:
> Looks cool, what about hiding just the cli for `--firefox.env` until its usable = next Geckodriver or will it be released soon do you know? I wanna get out the current things in master asap, but I guess that probably would be next week.

Can we hide the `--help` but keep the functionality?  I expect the next geckodriver release to be longer than a week for sure.

soulgalore:
Yep let me merge and hide :)

PR #1111, opened 2020-01-14
Add iteration set delay support

SamuliR:
used for --browsertime.iterationSetDelay

### DIFF
--- a/lib/core/engine/index.js
+++ b/lib/core/engine/index.js
@@ -48,6 +48,10 @@ class Engine {
     this.myExtensionServer = new ExtensionServer();
   }
 
+  async sleep(ms) {
+    return new Promise(resolve => setTimeout(resolve, ms));
+  }
+
   /**
    * Start the engine. Will prepare everything before you will start your run:
    * * Start XVFB (if it is configured)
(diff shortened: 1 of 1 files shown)


soulgalore:
Hi @SamuliR is this the same as https://github.com/sitespeedio/browsertime/blob/master/lib/core/engine/index.js#L132-L134 ? I didn't fully get that. 

Best
Peter

SamuliR:
iterationSetDelay is the delay in-between a set of iterations so it waits a little bit before the next bombardment

iteration - delay - iteration - delay - iteration 
**iterationSetDelay** 

iteration - delay - iteration - delay - iteration
**iterationSetDelay** 

and so on

matti:
@soulgalore so that you can run 10 iterations with normal delay, for example, 10s and then sleep 500s.

effectively: test 10 iterations, sleep 500s, test 10 iterations... currently this is impossible - we have successfully run monitoring tests like this so that we get more samples per time and then pause.

soulgalore:
Yeah but do you really need that in browsertime or how do you run it?

```
browsertime ...
sleep 10 
browsertime ...
```

Or is the use case you test multiple URLs from sitespeed.io? Then I think we could ad that directly in sitespeed.io so we have an extra delay/wait before moving on to next URL.

matti:
when running with sitespeed with docker (the pre-built sitespeed image, not browsertime) it would be super handy!

soulgalore:
Yep let me add it before the next release.

PR #1114, opened 2020-01-16
Fix ContentfulSpeedIndex - fixes #1091

### LINKED ISSUE #1091: Contentful Speed Index's visualmetrics.py broken on convert call on macOS
when visualmetrics is processing the data for the contentful speed index it breaks
with all versions of ImageMagick (6.x, 7.x) because it fails to parse the convert result.

For example, this call:

```
convert browsertime-results/google.com/2020-01-09T143331+0100/pages/google.com/data/video/images/3/ms_001400.png -canny 2x2+8%+8% -define histogram:unique-colors=true -format %c histogram:info:-'
```

returns 

```
     80883: (0,0,0) #000000 gray(0)
       317: (65535,65535,65535) #FFFFFF gray(255)
```

and that leads to 

```
Traceback (most recent call last):
  File "/Users/tarek/Dev/gecko/mozilla-central-opt/tools/browsertime/node_modules/browsertime/vendor/visualmetrics.py", line 1909, in main
    options.herodata)
  File "/Users/tarek/Dev/gecko/mozilla-central-opt/tools/browsertime/node_modules/browsertime/vendor/visualmetrics.py", line 1330, in calculate_visual_metrics
    value, value_progress = calculate_contentful_speed_index(progress, dirs)
  File "/Users/tarek/Dev/gecko/mozilla-central-opt/tools/browsertime/node_modules/browsertime/vendor/visualmetrics.py", line 1493, in calculate_contentful_speed_index
    value  = int(output.split()[7].split(':')[0])
ValueError: invalid literal for int() with base 10: 'gray(255)'
```

@soulgalore  Would it make sense to extract X from gray(X) via a regexp here ?
Happy to do this.

tarekziade:


soulgalore:
Thank you @tarekziade !

soulgalore:
Ooops sorry I didn't test this locally, this totally blew up on our test server that test against a couple of URLs with the latest commit, so I reverted. I will have a look at this later today or tomorrow. Building a docker container also generated errors when running so I think something is wrong somewhere in the commit :)

tarekziade:
@soulgalore  sorry about this regressiin... it worked on my tests. does your test server do something specific ? Can I run the same thing on my side to make sure everything is fine?

soulgalore:
No worries @tarekziade the same happens to me all the time, that's why I've the test setup to test master :)

So we got the same error on Travis but we swallow the error. In the past we have had so many error log messages on Travis but I think most is fixed now and we (I) should make sure we exit with a non 0 so travis can signal the error.

See https://travis-ci.org/sitespeedio/browsertime/builds/637841655?utm_source=github_status&utm_medium=notification
```
08:25:41.434 - Could not find the contentfulness value
178308:25:41.434 - 'int' object is not iterable
1784Traceback (most recent call last):
1785  File "/home/travis/build/sitespeedio/browsertime/vendor/visualmetrics.py", line 1936, in main
1786    options.herodata)
1787  File "/home/travis/build/sitespeedio/browsertime/vendor/visualmetrics.py", line 1330, in calculate_visual_metrics
1788    value, value_progress = calculate_contentful_speed_index(progress, dirs)
1789TypeError: 'int' object is not iterable
```

I made (a non merged) fix so we return correctly see https://github.com/sitespeedio/browsertime/pull/1115 but I guess this is something else?

tarekziade:
Oh ok I see -- I did a mismatch between the old (int value) and the new return value. 

I think it would be wise to add a few unit test for visualmetrics.py itself where we provide a few data samples to make sure we extract the right values for convert calls. I can rework the patch and add this.

tarekziade:
@soulgalore in the interim I found the issue

```
>>> output = """\
... 100696: (  0,  0,  0) #000000 gray(0)
...        504: (255,255,255) #FFFFFF gray(255)
... """
>>> import re
>>> lines = [line.strip() for line in output.split("\n") if line.strip()]
>>> lines
['100696: (  0,  0,  0) #000000 gray(0)', '504: (255,255,255) #FFFFFF gray(255)']
>>> matcher = re.compile(r'\d+: \S+ #[0-9A-F]+ (?:gray\((\d+)\)|(\d+)(?:))')
>>> match = [[v for v in el if v] for el in matcher.findall(lines[-1])]
>>> match
[['255']]
>>> int(match[0][0])
255
```
should be `matcher.findall(lines[-1])]` not `matcher.findall(lines[0])]`

Also, see the space in "(  0,  0,  0)" ? we'll need to make the regexp better at this next even though the second line is not affected

tarekziade:
hey @soulgalore do you have an irc channel or some place for chats ? could be easier to interact :)

soulgalore:
@tarekziade yes join https://sitespeedio.herokuapp.com if you use Slack!

PR #1115, opened 2020-01-16
If contentful fail, return none for both metric and progress

soulgalore:


tarekziade:
see also https://github.com/sitespeedio/browsertime/pull/1114#issuecomment-575197102

PR #1118, opened 2020-01-16
Rehabilitate `--processStartTime` on Android.

ncalexan:
This still isn't as robust as we'd like, but it works on the three Android devices that Mozilla cares about right now -- x86 emulator, Pixel 2, Moto G5 -- so it would be helpful to have it landed.


soulgalore:
@ncalexan perfectly fine for you to self merge or let one of your team mate merge things for Firefox if you see things that doesn't work or things that you feel needs to be changed.

PR #1121, opened 2020-01-17
fixed contentfulspeedindex and added a first unittest

tarekziade:
This patch fixes the content full speed index convert calls, and adds a first unittest.

The unitest is executed in Travis using pytest.

The patch also makes visualmetrics.py work on both python 2.7 and 3.x (at least for importing it and running calculate_contentful_speed_index() )


tarekziade:
This also makes visualmetrics.py work on both python 2.7 and 3.x (at least for importing it and running calculate_contentful_speed_index()

soulgalore:
Thank you @tarekziade !

Issue #1093, opened 2020-01-10
newStart for doing removeOrange is not correct. 


sefeng211:
I think the `newStart` variable defined [here](https://github.com/sitespeedio/browsertime/blob/master/lib/video/postprocessing/finetune/index.js#L21) needs to be adjusted. 

I did a few tests and this was what I saw
newStart is `1.723` and video starts at `1.722400`
newStart is `1.813` and video starts at `1.812278`

And later when we execute the ffmpeg command which cuts the video to remove the orange frame,  ~~an extra orange will be created, which causes the start of the output video is one frame off.~~. Some miss alignment happened to the frames, which caused the output became one frame off

soulgalore:
Hi @sefeng211 aha thanks. Do you have time to give it a try to fix it? Else I can have a look later on, I'm in the middle of verifying that all base functionality is working, when I'm done with that I can look at other things.

sefeng211:
Sure, I can try to fix it. In the meantime, I am also investigating other issues such as why the timestamp-overlay added video misses some frames (passing `-vsync 0` seems fix, however the timestamp overlay doesn't match the timestamp of the video)

sefeng211:
Closing this off as the patch is merged.

PR #1122, opened 2020-01-17
Fix the bug where a frame is missed after ffmpeg transformation

sefeng211:
This is for issue #1093 

We should return the ceiling value of the timestamp.

I did some testings, and I saw visual metrics numbers stayed the same as without the patch, so this was good. 

Some side effects may be things like `trimming frames from the end`, a frame which was excluded may be included after the patch, however I don't think this is a big problem.

### DIFF
--- a/vendor/visualmetrics.py
+++ b/vendor/visualmetrics.py
@@ -143,7 +143,7 @@ def extract_frames(video, directory, full_resolution, viewport):
             match = re.search(pattern, line)
             if match:
                 frame_count += 1
-                frame_time = int(math.ceil(float( \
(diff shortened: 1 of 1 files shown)


soulgalore:
It seems that pytest is missing, strange?

tarekziade:
weird... it's installed earlier with 

```
python -m pip install --user pytest
```

and it worked in previous branches/ maybe a path issues?

we can also replace that install with "apt install python-pytest"

soulgalore:
Yeah very strange. Let us merge this anyway and then fix it in master.

soulgalore:
Thank you @sefeng211

Issue #1123, opened 2020-01-17
Consider normalizing videos from VFR to CFR before any video transformation. 


sefeng211:
Android adb screen recorder records the video in variable frame rates and users are not allowed to adjust. 

Using ffmpeg to do operations on VFR videos would mess the timestamps up. 

I think we should consider to always convert VFR videos to CFR (60 FPS for example) before any video transformation. For example, before removing the orange frame. This also fixes #1108

soulgalore:
@sefeng211 you are right., We actually used to convert it to 60fps but is seems to have disappeared through the year, let me dig into the history. I'll ping you when I find it and consult you if it is the right way to do it!

soulgalore:
I removed it in 7737807ca1f67b10b2fe12bd0037d244ad9b728d but no good comment. I can re-add it this weekend, @sefeng211 do the old convert look good or did you think of something else?

sefeng211:
@soulgalore Thanks, it looks good :)

sefeng211:
Oh one thing is, this old conversion does the conversion right away, even before calculating visual metrics. I was thinking still to use the original video(VFR) for visual metrics calculation, and we convert it after that.

sefeng211:
It does change the visual metrics number a little bit, not a lot though, just a tiny bit of changes. However, since there are no evidence that shows the visual metrics numbers are wrong, I think we should keep using the original video to do visual comparisons, and convert it afterwards. What do you think @soulgalore ?

soulgalore:
Let me do branch tonight and we can have look what it looks like!

soulgalore:
Added a PR, but it seems it still off, I could see that first visual change metric isn't added exactly when it happens in the video.

sefeng211:
Hmm, Let me try to reproduce it.

sefeng211:
@soulgalore First Visual Change seems correct me. What was the issue?

soulgalore:
This was fixed right?

PR #1134, opened 2020-01-20
Convert the Android video to 60 fps before adding metrics/timer #1123

soulgalore:


soulgalore:
@sefeng211 yep I was wrong, it looks good, its only when we add the text that's not perfect in sync but that is something we can live with for now.

PR #1137, opened 2020-01-21
Refactor the code to only start the extension server when needed

soulgalore:
Lets do follow up PR where we add start/stop the server on demand and close before we navigate to the next page (as we do now in scripting).


soulgalore:
ping @gmierz think this is interesting for you

PR #1151, opened 2020-01-25
Use PageLoadStrategy none as default with higher default wait times.

soulgalore:
Using load strategy none and wait before doing the first check (and
the continous checks) decreases communication between the driver
and the browser.


ncalexan:
@gmierz this is relevant to one of the things we talked about, namely delaying on the host side, not the target side.

soulgalore:
yes, we discussed that before Christmas. I've been running that config for a month for Wikipedia and works fine there. For Chrome I could see a big change in stability but not for Firefox. But with the fix you did for getting the Geckodriver log, I could see that it is silent now when we are waiting on the first run of the complete check, so that is good!

Issue #1157, opened 2020-01-31
Handle "bare" domain names in new URL retry logic


ncalexan:
In Gecko-based browsers, the URL comparisons added by #1077 are failing because the similar looking URLs for "bare" domain names like `https://x.com` and `https://x.com/` do not string-compare as `===`.  It's probably not worth fully normalizing URLs (i.e., sorting query parameters) but it is worth normalizing such cases.

ncalexan:
This was merged.

PR #1158, opened 2020-01-31
Compare "bare" domain names in URL retry logic as equal. (#1157)

ncalexan:
This fixes an issue introduced in #1077: URL comparisons are failing
because the similar looking URLs for "bare" domain names like
https://x.com and https://x.com/ do not string-compare as ===.  It's
probably not worth fully normalizing URLs (i.e., sorting query
parameters) but it is worth normalizing trailing slashes in such
cases, which we do by parsing with `new URL(...)` (which is present
since Node v8).


soulgalore:
Great @ncalexan I've missed that.

PR #1165, opened 2020-02-06
Record orange frame duration in browsertime

sefeng211:
videoRecordingStart is the timestamp of the first frame that's is not orange frame, which indicates how long the orange frame in video.

Let's record it so that we can measure it.

@soulgalore let me know if you have any objections :)


soulgalore:
Lets push this now and if there's an issue, we can fix that later :)

PR #1166, opened 2020-02-06
Add a temporary workaround for orange frame removal issue

sefeng211:
Visual metrics are not correct when running browsertime
with the new Firefox for Android browser (Fenix) when
WebRender is enabled, because there's a bug in WebRender.

This is tracking bug in bugzilla:
 - https://bugzilla.mozilla.org/show_bug.cgi?id=1606365

In an ideal world, we fix the bug in WebRender, however I
don't know when that will happen, so this is the workaround.


sefeng211:
@soulgalore agreed! How does it look now?

soulgalore:
Looks great, thank you @sefeng211  for the change :)

PR #1167, opened 2020-02-06
Add new resource related performance metrics

sefeng211:
These two numbers help us to compare resource differences between Browsers. 

@soulgalore Please let me know if you have objections :)

### DIFF
--- a/browserscripts/pageinfo/resources.js
+++ b/browserscripts/pageinfo/resources.js
@@ -1,0 +1,14 @@
+(function() {
+  const resources = window.performance.getEntriesByType("resource");
+
+  let resourceDuration = 0;
+  for (let i = 0; i < resources.length; i++) {
+    resourceDuration += resources[i].duration;
+  }
+
+  return {
+    resourceCount: Number(resources.length),
+    resourceDuration: Number(resourceDuration)
+  }
+})();
+


browserscripts/timings/pageTimings.js
```diff
@@ -14,7 +21,9 @@
       domContentLoadedTime: Number(t.domContentLoadedEventStart.toFixed(d)),
       pageLoadTime: Number(t.loadEventStart.toFixed(d)),
       frontEndTime: Number((t.loadEventStart - t.responseEnd).toFixed(d)),
-      backEndTime: Number(t.responseStart.toFixed(d))
+      backEndTime: Number(t.responseStart.toFixed(d)),
+      resourceCount: Number(resources.length),
```
soulgalore: Do you need to have them all in this JSON result or can you easily use the others? Thinking resource count could be moved to /pageinfo/ and named numberOfResources or if you create a new file named resources.js and with count and duration? We used to have all resource timings but it was generating so much data.

sefeng211: I can move it to `resources.js` and that's fine. Would you be able to find the resource timings that we used to have? I am curious

soulgalore: @sefeng211 https://github.com/sitespeedio/browsertime/commit/3fbc5053752fbf93c5062d75e92048102692a9a7#diff-515ed627ef20d628635e7a99d6f0debe

The produced JSON was unreadable when we get that amount of data in the result JSON, so I think your fix is much better when we get that into another file :)

sefeng211: let me know how it looks :)

soulgalore: Cool, I'll merge it after I've done a fix release. I will just rename the metric to duration and count, since the data structure will be taken from  the file name:
```
resources: { duration: XXX,
count: YYY
}
```

soulgalore:
FYI @sefeng211 you can also add your own JS to collect the metrics you need from your test using `--script` but merging this since its useful for everyone running without getting a HAR (also Safari at the moment).

PR #1182, opened 2020-02-10
Fix ContentfulSpeedIndex and VisualMetrics Tests 

### LINKED ISSUE #1178: ContentfulSpeedIndex is not correct
There's bug in how we use the [`matcher` regex](https://github.com/sitespeedio/browsertime/blob/master/browsertime/visualmetrics.py#L1680) I think.

Say the output of the `convert` command is `'534: (123,456,789) #FFFFFF gray(444)'`. In this case, we want the value to be `534` which is the pixel count I think, but the `matcher` regex would return `444` from it. See https://regex101.com/r/C1p0Pg/1. 

This also indicates the ContentfulSpeedIndex test is not correct or sufficient. 

We should fix it and it's not clear to me of why we complicate it with this regex, actually this entire line feels a bit overkill to me.

sefeng211:
To fix #1178 

There are 2 issues in `test_visualmetrics.py`
1) `assertTrue` just verifies the trueness of the values, and we want to
   compare the values.
2) progress should be sorted based on the timestamps of the images.

I also modified the regex to get the digits before the `:`, and then we pick the last one as the value. I believe this is platform-independent. @tarekziade May I ask you to test on MacOS? I don't have a mac machine :)

Thanks!

### DIFF
--- a/browsertime/test_visualmetrics.py
+++ b/browsertime/test_visualmetrics.py
@@ -20,8 +20,10 @@ class TestVisualMetrics(unittest.TestCase):
             return p
 
         progress = [_p(image) for image in images if image.startswith("ms_")]
-        res = calculate_contentful_speed_index(progress, directory)
-        self.assertTrue(res[0], 5080)
+        res = calculate_contentful_speed_index(sorted(progress,
+                                                      key = lambda image: image['time']),
+                                               directory)
+        self.assertEqual(res[0], 5080)
 
     def test_calculate_perceptual_speed_index(self):
         directory = "test_data"
@@ -33,5 +35,7 @@ class TestVisualMetrics(unittest.TestCase):
             return p
 
         progress = [_p(image) for image in images if image.startswith("ms_")]
-        res = calculate_perceptual_speed_index(progress, directory)
-        self.assertTrue(res[0], 5080)
+        res = calculate_perceptual_speed_index(sorted(progress,
+                                                      key = lambda image: image['time']),
+                                               directory)
+        self.assertEqual(res[0], 946)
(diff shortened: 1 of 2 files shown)


soulgalore:
FYI I'm working on getting Travis tests to run on OS X too, hopefully I get it to work in a couple of days.

browsertime/test_visualmetrics.py
```diff
@@ -20,8 +20,10 @@ def _p(image):
             return p
 
         progress = [_p(image) for image in images if image.startswith("ms_")]
```
tarekziade: maybe we can factor this progress var out since it's used in both tests.

sefeng211:
@soulgalore Look like the travis CI error I got were caused by the browsertime we used the in CI is a old version, which doesn't have changes I made in https://github.com/sitespeedio/browsertime/pull/1122/files. 

So I'd like to merge the PR despite the tests are failing, objections?

soulgalore:
@sefeng211 any reason you can't update your branch with the latest in master or how do you mean? :) Feel free to merge if you feel sure that's the reason.

sefeng211:
@soulgalore Correct me if I am wrong. The way I see how the test works is that, it uses browsertime as a package and then import it into the test. Do you think it could use a different version than the current branch? I am a bit confused about the failure, the only possible explanation I see is it used a different browsertime.visualmetrics.

soulgalore:
Hi @sefeng211 no it runs the local file right see https://github.com/sitespeedio/browsertime/blob/master/lib/video/postprocessing/visualmetrics/visualMetricsPath.js?

sefeng211:
I found my local and CI produce different `convert` results. I see they uses a different version, so I updated the test case to obey the CI :)

soulgalore:
Aha @sefeng211 good fine (but scary).

PR #1185, opened 2020-02-10
Use --firefox.profileTemplate setting when it's provided.

gmierz:
This patch fixes an issue where the profile that is provided to `--firefox.profileTemplate` is not used. With this change, if there exists a setting for `--firefox.profileTemplate` it will be used, otherwise it defaults to the profile in `browsertime/browsersupport/firefox-profile`.


soulgalore:
thank you @gmierz !

PR #1186, opened 2020-02-11
Add win32 to x11recorder.js origin/offset options.

gmierz:
This patch fixes an issue where the origin and offset for the `x11recorder` are undefined on windows machines. Note that `process.platform` returns win32 for any type of windows machine.

### DIFF
--- a/lib/video/screenRecording/desktop/x11recorder.js
+++ b/lib/video/screenRecording/desktop/x11recorder.js
@@ -16,7 +16,7 @@ const screenOffsets = {
   windows: {
     firefox: { origin: '0,71', offset: { x: 0, y: 168 } },
     chrome: { origin: '0,66', offset: { x: 0, y: 66 } } 
-  }
+  },
   darwin: {
     firefox: { origin: '0,71', offset: { x: 0, y: 80 } },
     chrome: { origin: '0,66', offset: { x: 0, y: 80 } },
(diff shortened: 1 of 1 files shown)


soulgalore:
Ok, @gmierz do you know if it works for Windows or do we need some FFMPEG config changes?

gmierz:
@soulgalore I'm not sure if we need some FFMPEG config changes. This fixes an issue that occurs when there is no screen recording being done. The speedometer tests in this test run have this failure on initialization: https://treeherder.mozilla.org/#/jobs?repo=try&tier=1%2C2%2C3&revision=5573be3eb3a69efa9a88978a9fc84ec94621af59&selectedJob=288378127

There are other tests there that record video, but they use the Firefox window recorder. I haven't been able to get FFMPEG recording working on Windows (locally) though because of the error `Unknown input format: 'x11grab'`.

soulgalore:
Ah I see @gmierz! I think there's some experimental Windows on Travis, so I can see if I can get a test up and running. 

> ffmpeg

There are different captures for OSX/Linux/Windows so x11grap will not work on Windows, we need to change to gdigrab (checkout https://trac.ffmpeg.org/wiki/Capture/Desktop), something like the changes I added for OS X.

gmierz:
Thanks for merging!

Oh cool, thanks for letting me know about that `gdigrab` option - I'll play with that a bit the next chance I get. :)

PR #1202, opened 2020-02-17
Support video recording on Windows

soulgalore:


soulgalore:
This fixes recording as far as I can see but there's a couple of more things that fails:
* Running Chrome we get the popup: https://stackoverflow.com/questions/23087724/chromedriver-disable-developer-mode-extensions-pop-up-on-selenium-webdriver-au/44636414
* Installing imagemagick through choco doesn't include mogrify, installing it and choosing legacy commands fixes that
* Still get errors from VisualMetrics `mogrify.exe: unable to open image '...peed.io\data\filmstrip\1\ms_*.png': Invalid argument @ error/blob.c/OpenBlob/3496.`

PR #1204, opened 2020-02-18
Use %% for percentages in argparser help

brennie:
The help strings of `argparser.ArgumentParser` arguments are interpreted
as `%-formatting` strings so strings like "% o" are interpreted as "%o"
(i.e., format as octal). Help text has been corrected to use `%%`, which
generates a single `%` in output.


soulgalore:
Cool, thanks @brennie !

Issue #1209, opened 2020-02-26
vis-colors temporary directory is never removed


dpalmeiro:
After it is finished, visualmetrics.py does not clean up the vis-colors temporary directory.

PR #1210, opened 2020-02-26
Remove vis-color temp dir during cleanup

dpalmeiro:
Fixes https://github.com/sitespeedio/browsertime/issues/1209


tarekziade:
Semi-related, we create this tempdir even if it's not used, we could probably do a better job there

soulgalore:
Thanks @dpalmeiro !

Issue #1216, opened 2020-03-09
single module test script


tarekziade:
I'd like to have the ability to pass to Browsertime a single JS module containing postScript, preScript and tests functions to make it look closer to a standalone test script a developer can have in their tests suite.

current proposed template:

```
    async function setUp(context, commands) {
      console.log("setUp");
    };
    ​
    async function perfTest1(context, commands) {
      return await commands.measure.start('https://www.sitespeed.io/examples/');
    };
    ​
    async function perfTest2(context, commands) {
      return await commands.measure.start('https://www.sitespeed.io/examples/');
    };
    ​
    async function tearDown(context, commands) {
      console.log("tearDown");
    };
 
    module.exports = {
      setUp: setUp,
      tearDown: tearDown,
      tests: [perfTest1, perfTest2]
    };
```

when this script is loaded, BT will look at the three variables exported by the module and use them, or fallback to the existing single function model.

soulgalore:
Hi @tarekziade cool. Let me think about it for a couple of days. Thinking maybe we could just use the current scripting with some small fixes, but that's maybe what you already done in your patch, I haven't looked.

tarekziade:
yeah, let me update the runner so you can see what I've done. of course this is just the prototype, I'll need to do it in a cleaner way with tests etc

PR #1217, opened 2020-03-09
minimal patch for #1216

tarekziade:


tarekziade:
Example of script that can be used with this patch:

```
async function setUp(context, commands) {
  console.log("setUp");
};

async function perfTest1(context, commands) {
  return await commands.measure.start('https://www.sitespeed.io/examples/');
};

async function perfTest2(context, commands) {
  return await commands.measure.start('https://www.sitespeed.io/examples/');
};

async function tearDown(context, commands) {
  console.log("tearDown");
};

module.exports = {
  setUp: setUp,
  tearDown: tearDown,
  tests: [perfTest1, perfTest2]
};
```

launched with:

```
$ browsertime.js -b firefox --skipHar --iterations 1 test_basic.js
```

soulgalore:
Hi @tarekziade yeah this can be cool. I'm thinking we could add alias for preScript and postScript and use your standard setUp/tearDown too, that would be nice. Then we can slowly move to that standard, that makes more sense I think.

Can you explain a little more of your use case of having multiple tests in one file (perfTest1, perfTest2)? My reason to been able to handle multiple script files before was to be able separate use cases into multiple files.

tarekziade:
> Can you explain a little more of your use case of having multiple tests in one file (perfTest1, perfTest2)? My reason to been able to handle multiple script files before was to be able separate use cases into multiple files.

@soulgalore ok. 

My use case is just organizational (have the ability to reunite several performance tests in one single JS file), but I realize now the patch won't do exactly what should be done since from BT point of view, they will be reunited in a single set of results. 

I also realise that this might be a small shift into the "page load test" paradigm.

The idea with perfTest1/perfTest2 is to let a developer set up the browser for doing a series
of **distinct** performance tests.

Example: I want to log in to my FxA account and then do two performance tests on the browser -- two different scenari, then disconnect on tearDown. 

But maybe that's a bit too different from the current mechanism. It would also involve adding a new option if we want to select a single perftest in a script.

soulgalore:
So I'm thinking that is almost what we are doing in sitespeed.io. So you could create your own wrapper like https://github.com/sitespeedio/sitespeed.io/tree/master/lib/plugins/browsertime and then you could do that slice on the outside. Maybe there's some fixes that needs to be done on how we store assets. I'll ping you so we can talk.

tarekziade:
replaced by https://github.com/sitespeedio/browsertime/pull/1220

PR #1220, opened 2020-03-18
Added supported for unified scripts - fixes #1216

### LINKED ISSUE #1216: single module test script

tarekziade:


.travis.yml
```diff
@@ -43,6 +43,7 @@ matrix:
         - node ./bin/browsertime.js -b firefox test/navigationscript/multi.js https://www.sitespeed.io/blog/ -n 1
         - node ./bin/browsertime.js -b chrome test/navigationscript/multi.js -n 3 --chrome.timeline --video --visualMetrics --visualElements --viewPort 1000x600
         - node ./bin/browsertime.js -b firefox --firefox.geckoProfiler -n 1 https://www.sitespeed.io/ --spa
+        - node ./bin/browsertime.js -b firefox test/unifiedscripts/sample.js
```
soulgalore: Thinking it is ok to run on one OS, since if its broken its 99,99% broken on all platforms? I haven't documented it (sorry) but the other tests on OS X and Windows is mostly to verify browsers that only works on that platform or simple generic tests.

bin/browsertime.js
```diff
@@ -94,6 +101,44 @@ async function run(urls, options) {
...
+/*
+  Each url can be:
+   - an url value
+   - an array of tests. In that case it's a mapping containing
+     theses values:
+     - test: an async function containing the test to run
+     - setUp: an async function for the preScript [optional]
+     - tearDown: an async function for the postScript [optional]
+*/
+var tests = [];
```
soulgalore: const is the new var :) Meaning in our code you can almost always use const (or let). I should document that too, sorry.

soulgalore:
Cool thank you @tarekziade !

PR #1227, opened 2020-03-26
[wip] Fix CSI computation

### LINKED ISSUE #1226: ContentfulSpeedIndex (CSI) is currently problematic because it punishes staying on the last frame if the last frame is not the most contentful
The [CSI algorithm implemented in visualmetrics.py](https://github.com/sitespeedio/browsertime/blob/0d01f12f2a8eb2c6d8c935f44b8c28fcd37fe33b/browsertime/visualmetrics.py#L1649-L1705) currently does the following:
 1. For each frame, calculate a contentfulness score (based on edge detection and counting "edge" pixels).
 2. Find the maximum contentfulness across all frames.
 3. For each frame, calculate the percentage by which the frame differs from the maximum contentfulness.
 4. Add up the differences weighted by time.

This works ok if the last frame is the most contentful frame. However, if maximum contentfulness was reached before the final state, then the final state gets puninshed: The longer the video stays in the final state, the higher (worse) the score becomes, because it's not at its most contentful during that time.
The duration of the final state in the captured videos is somewhat arbitrary. We rely somewhat on the assumption that some extra video seconds in the final state don't affect the score - but for CSI they do.

I have an example of a Fenix and a Chrome recording where Chrome puts content on the screen 200ms before Fenix. But the CSI score for Chrome is worse (higher), because it briefly shows some extra content (a gray line at the right edge of the viewport) that then goes away again, and this extra content punishes all remaining frames that don't have it.

Proposal: Rather than computing the maximum contentfulness, I think we should compare all frames to the "final" contentfulness. If a frame in the middle has more contentfulness than the final frame, then that frame should be treated as if it had the same contentfulness as the final frame. This rewards putting content on the screen early, and it doesn't punish staying in the final state longer.

mstange:
This fix is completely untested. Fixes #1226.

### DIFF
--- a/browsertime/visualmetrics.py
+++ b/browsertime/visualmetrics.py
@@ -1649,8 +1649,8 @@ def calculate_speed_index(progress):
 def calculate_contentful_speed_index(progress, directory):
     try:
         dir = os.path.join(os.path.dirname(os.path.abspath(__file__)), directory)
-        def get_filename(p):
-            return os.path.join(dir, "ms_{0:06d}.png".format(p["time"]));
+        def get_filename(t):
+            return os.path.join(dir, "ms_{0:06d}.png".format(t));
 
         matcher = re.compile(r"(\d+?):")
         def compute_contentfulness(frame):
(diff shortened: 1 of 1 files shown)


soulgalore:
Hi @mstange I've changed to use **main** as default branch. When you all have had time to look at the CSI computations, please make another PR. Thanks!!

PR #1234, opened 2020-04-14
Fix checks to work with Python 3.

soulgalore:


soulgalore:
The failing test is unrelated (let me update the container tomorrow).

PR #1246, opened 2020-04-20
fix typo in an error message of click.js when a css selector couldn't be clicked for some reason

petemyron:


soulgalore:
Thank you @petemyron !

PR #1248, opened 2020-04-21
set orange.style.pointerEvents = 'none' 

Icecold777:
Mouse focus on element won't be lost at script measurement start.
Relates to sitespeed.io issue
https://github.com/sitespeedio/sitespeed.io/issues/2956


soulgalore:
Thank you @Icecold777 !

PR #1252, opened 2020-04-23
Add --firefox.appconstants flag to obtain AppConstants information.

gmierz:
This patch removes the AppConstants/asyncAppConstants from the results JSON. They can still be obtained by supplying the  `--firefox.appconstants` option.

### DIFF
--- a/lib/firefox/webdriver/firefoxDelegate.js
+++ b/lib/firefox/webdriver/firefoxDelegate.js
@@ -214,8 +214,13 @@ class FirefoxDelegate {
   }
 
   async afterCollect(index, result) {
-    if (this.firefoxConfig.geckoProfiler && this.options.visualMetrics) {
-      for (let i = 0; i < result.length; i++) {
+    for (let i = 0; i < result.length; i++) {
+      if (!this.options.firefox.appconstants) {
+        delete result[i].browserScripts.browser.appConstants;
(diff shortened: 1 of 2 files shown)


gmierz:
@soulgalore This patch solves the issue you raised, but I'm not sure if it's the ideal solution here. Let me know if you think this should be implemented differently.

soulgalore:
Thanks @gmierz for fixing it! This works fine.

There's one other way to solve it: You can feed configurations into a script in a generic way, but I don't think its documented. Checkout https://github.com/sitespeedio/browsertime/blob/883c692db497b7213d6f9f9e0c67ea59544b71db/lib/core/seleniumRunner.js#L522-L523

With visual elements as an example: https://github.com/sitespeedio/browsertime/blob/883c692db497b7213d6f9f9e0c67ea59544b71db/browserscripts/pageinfo/visualElements.js
and https://github.com/sitespeedio/browsertime/blob/883c692db497b7213d6f9f9e0c67ea59544b71db/lib/support/cli.js#L505-L508

So automagically if you add  `--scriptInput.appConstants somethingsomething ` browsertime will pass that on to your script if you make sure it is passed into your script following the pattern of visual elements. But I don't remember if its support the async scripts, so just deletes it works great now.

I'll merger it ASAP before I push the next release.

soulgalore:
Thank you @gmierz !

Issue #1211, opened 2020-02-26
Provide an option to re-enable Firefox Tracking Protection


acreskeyMoz:
By default, Browsertime disables Firefox's Tracking Protection feature and supporting tech.

See: 
[firefoxPreferences.js](https://github.com/sitespeedio/browsertime/blob/52121ec7af3583987ba1ec58ba8796e49762b7c6/lib/firefox/webdriver/firefoxPreferences.js#L167-L168)

Tracking Protection can have a significant impact on performance (particularly 'Strict' mode which blocks resources from identified trackers.)

This bug is to provide an browsertime option so that Tracking Protection can be enabled if desired.

There will be some work to determine the complete set of preferences from `firefoxPreferences.js` which will need to be optionally removed. (The underlying tech relies on SafeBrowsing).

As a first attempt, I believe the following would be sufficient:

```
'browser.safebrowsing.blockedURIs.enabled': false,
  'browser.safebrowsing.downloads.enabled': false,
  'browser.safebrowsing.downloads.remote.url':
    'http://127.0.0.1/safebrowsing-dummy/downloads',
  'browser.safebrowsing.malware.enabled': false,
  'browser.safebrowsing.passwords.enabled': false,
  'browser.safebrowsing.phishing.enabled': false,
  'browser.safebrowsing.provider.google.gethashURL':
    'http://127.0.0.1/safebrowsing-dummy/gethash',
  'browser.safebrowsing.provider.google.updateURL':
    'http://127.0.0.1/safebrowsing-dummy/update',
  'browser.safebrowsing.provider.google4.gethashURL':
    'http://127.0.0.1/safebrowsing4-dummy/gethash',
  'browser.safebrowsing.provider.google4.updateURL':
    'http://127.0.0.1/safebrowsing4-dummy/update',
  'browser.safebrowsing.provider.mozilla.gethashURL':
    'http://127.0.0.1/safebrowsing-dummy/gethash',
  'browser.safebrowsing.provider.mozilla.updateURL':
    'http://127.0.0.1/safebrowsing-dummy/update',

... 
'privacy.trackingprotection.annotate_channels': false,
  'privacy.trackingprotection.enabled': false,
```

soulgalore:
@acreskeyMoz the reason for this was that the Tracking Protection introduced so much variance in the metrics, since when running a new profile every time, it download the tracking protection lists during the page that is under test, do you know if that has been changed?

acreskeyMoz:
[...]

PR #1272, opened 2020-05-06
Make it possible to enable Safe Browsing and Tracking protection.

soulgalore:
Lets do this in multiple steps. First make it possible to enable them, then
create a profile where all lists are already downloaded and then in next
major enable it by default.

https://github.com/sitespeedio/browsertime/issues/1211 and @acreskeyMoz

### DIFF
--- a/lib/firefox/webdriver/firefoxPreferences.js
+++ b/lib/firefox/webdriver/firefoxPreferences.js
@@ -88,25 +88,6 @@ module.exports = {
   'browser.newtabpage.activity-stream.default.sites': '',
   'browser.newtabpage.activity-stream.telemetry': false,
   'browser.reader.detectedFirstArticle': true,
-  'browser.safebrowsing.blockedURIs.enabled': false,
-  'browser.safebrowsing.downloads.enabled': false,
-  'browser.safebrowsing.downloads.remote.url':
-    'http://127.0.0.1/safebrowsing-dummy/downloads',
-  'browser.safebrowsing.malware.enabled': false,
-  'browser.safebrowsing.passwords.enabled': false,
-  'browser.safebrowsing.phishing.enabled': false,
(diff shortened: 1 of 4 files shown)


lib/firefox/webdriver/index.js
```diff
@@ -9,6 +9,7 @@ const isEmpty = require('lodash.isempty');
 const get = require('lodash.get');
 const defaultFirefoxPreferences = require('./firefoxPreferences');
 const disableSafeBrowsingPreferences = require('./disableSafeBrowsingPreferences');
```
acreskeyMoz, line 11: Ok great, I didn't realize that we already had an option to disable SafeBrowsing.

lib/firefox/webdriver/firefoxPreferences.js
```diff
@@ -88,25 +88,6 @@ module.exports = {
...
-  'browser.safebrowsing.provider.google.updateURL':
-    'http://127.0.0.1/safebrowsing-dummy/update',
-  'browser.safebrowsing.provider.google4.gethashURL':
-    'http://127.0.0.1/safebrowsing4-dummy/gethash',
-  'browser.safebrowsing.provider.google4.updateURL':
-    'http://127.0.0.1/safebrowsing4-dummy/update',
-  'browser.safebrowsing.provider.mozilla.gethashURL':
-    'http://127.0.0.1/safebrowsing-dummy/gethash',
-  'browser.safebrowsing.provider.mozilla.updateURL':
-    'http://127.0.0.1/safebrowsing-dummy/update',
```
acreskeyMoz, line 109: Independent of tracking protection, it's good that this was removed since Browsertime already had an option for disabling safebrowsing.

soulgalore: Yeah I must have missed that when I merged.

PR #1296, opened 2020-06-01
Add support for Android power testing

gmierz:
This patch adds the ability to perform power usage testing with Firefox on Android phones.

@soulgalore this code is very similar to what we use in the mozilla-central tree but fine-tuned for browsertime. :)

I'm not entirely sure about the structure or how we should organize all of this within browsertime - the big thing I am wondering about is where should `parsePowerMetrics` go?

I've only enabled power testing with Firefox for the time being (but other browsers will be easy to add in the future).

### DIFF
--- a/lib/android/index.js
+++ b/lib/android/index.js
@@ -314,8 +314,74 @@ class Android {
   }
 }
 
+async function parsePowerMetrics(batterystats, packageName) {
+  const reUid = new RegExp(`.+proc=([^:]+):"${packageName}".*`);
+  const reEstimations = new RegExp(`.+Estimated power use [(]mAh[)].*`);
+  const reScreen = new RegExp(`.+Screen: ([0-9.]+).*`);
+  const reWifi = new RegExp(`.+Wifi: ([0-9.]+).*`);
+
+  let uid = null;
+  let foundEstimations = false;
+  let rePower = null;
+  let found = false;
+  let powerData = {};
+
+  batterystats.split(/\r?\n/).forEach(line => {
+    if (!uid) {
+      let match = line.match(reUid);
+      if (match) {
+        uid = match[1];
+        rePower = new RegExp(`.+Uid ${uid}: ([0-9.]+)(.*)`);
+      }
+      return;
+    }
+    if (!foundEstimations) {
+      let match = line.match(reEstimations);
+      if (match) {
+        foundEstimations = true;
+      }
+      return;
+    }
+    if (!("full-screen" in powerData)) {
+      let match = line.match(reScreen);
+      if (match){
+        powerData["full-screen"] = parseFloat(match[1]);
+      }
+    }
+    if (!("full-wifi" in powerData)) {
+      let match = line.match(reWifi);
+      if (match){
+        powerData["full-wifi"] = parseFloat(match[1]);
+      }
+    }
+    if (rePower && !found) {
+      let match = line.match(rePower);
+      if (!match) {
+        return;
+      }
+      found = true;
+      powerData["total"] = parseFloat(match[1]);
+
+      let breakdown = match[2];
+      if (breakdown) {
+        match = breakdown.match(/([a-zA-Z]+)=([0-9.]+)/g);
+        if (!match) {
+          return;
+        }
+        match.forEach(category => {
+          let [name, value] = category.split("=");
+          powerData[name] = parseFloat(value);
+        })
+      }
+    }
+  })
+
+  return powerData;
+}
+
 module.exports = {
   Android,
+  parsePowerMetrics,
   isAndroidConfigured(options) {
     if (options.android === true) {
       return true;
--- a/lib/android/index.js
+++ b/lib/android/index.js
@@ -342,16 +342,16 @@ async function parsePowerMetrics(batterystats, packageName) {
       }
       return;
     }
-    if (!("full-screen" in powerData)) {
+    if (!('full-screen' in powerData)) {
       let match = line.match(reScreen);
-      if (match){
-        powerData["full-screen"] = parseFloat(match[1]);
+      if (match) {
+        powerData['full-screen'] = parseFloat(match[1]);
       }
     }
-    if (!("full-wifi" in powerData)) {
+    if (!('full-wifi' in powerData)) {
       let match = line.match(reWifi);
-      if (match){
-        powerData["full-wifi"] = parseFloat(match[1]);
+      if (match) {
+        powerData['full-wifi'] = parseFloat(match[1]);
       }
     }
     if (rePower && !found) {
@@ -360,7 +360,7 @@ async function parsePowerMetrics(batterystats, packageName) {
         return;
       }
       found = true;
-      powerData["total"] = parseFloat(match[1]);
+      powerData['total'] = parseFloat(match[1]);
 
       let breakdown = match[2];
       if (breakdown) {
@@ -369,12 +369,12 @@ async function parsePowerMetrics(batterystats, packageName) {
           return;
         }
         match.forEach(category => {
-          let [name, value] = category.split("=");
+          let [name, value] = category.split('=');
           powerData[name] = parseFloat(value);
-        })
+        });
       }
     }
-  })
+  });
 
   return powerData;
 }
--- a/lib/android/index.js
+++ b/lib/android/index.js
@@ -27,6 +27,11 @@ class Android {
       get(options, 'firefox.android.deviceSerial')
     );
     this.port = options.devToolsPort;
+
+    // Variables for android power testing
+    this.screenBrightnessMode = 0;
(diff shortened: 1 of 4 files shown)


lib/firefox/webdriver/firefoxDelegate.js
```diff
@@ -23,12 +27,39 @@ class FirefoxDelegate {
...
+      await this.android._runCommand(
+        'settings put system screen_brightness_mode 0'
+      );
+
+      // Set screen brightness to 50%
+      this.screenBrightness = await this.android._runCommandAndGet(
+        'settings get system screen_brightness'
+      );
+      await this.android._runCommand(
+        'settings put system screen_brightness 127'
```
soulgalore, line 60: I think we could move all Android commands here as a utility method: https://github.com/sitespeedio/browsertime/blob/master/lib/android/index.js

Then it will be less code to implement it for Chrome and other browser (and easier to find). I think all `.android._runCommand(...)` could be moved to that Android utility file instead.

lib/firefox/webdriver/firefoxDelegate.js
```diff
@@ -23,12 +27,39 @@ class FirefoxDelegate {
       : 'none';
     this.firefoxConfig = options.firefox || {};
     this.options = options;
+    this.android = null;
```
soulgalore, line 30: Minor: I don't think you need to initialise it at all since we always check if `this.android` exists before we use it, right?

lib/android/index.js
```diff
@@ -314,8 +314,74 @@ class Android {
   }
 }
 
+async function parsePowerMetrics(batterystats, packageName) {
+  const reUid = new RegExp(`.+proc=([^:]+):"${packageName}".*`);
+  const reEstimations = new RegExp(`.+Estimated power use [(]mAh[)].*`);
+  const reScreen = new RegExp(`.+Screen: ([0-9.]+).*`);
+  const reWifi = new RegExp(`.+Wifi: ([0-9.]+).*`);
+
+  let uid = null;
```
soulgalore, line 323: Minor: for less code I think you could just write `let uid;`

lib/android/index.js
```diff
@@ -314,8 +314,74 @@ class Android {
...
+  let foundEstimations = false;
+  let rePower = null;
```
soulgalore, line 325: Same as ^ (you can skip null)

lib/firefox/webdriver/firefoxDelegate.js
```diff
@@ -23,12 +27,39 @@ class FirefoxDelegate {
...
     this.hars = [];
+    if (isAndroidConfigured(this.options)) {
+      this.android = new Android(this.options);
+    }
+
+    if (this.android && this.options.androidPower) {
+      log.info('Initializing power testing');
+
+      // Disable adaptive brightness
+      this.screenBrightnessMode = await this.android._runCommandAndGet(
```
soulgalore, line 48: I'm wondering maybe this is something we always wanna do, even  when we don't run the power function to make sure all tests run with the same settings?

lib/firefox/webdriver/firefoxDelegate.js
```diff
@@ -38,6 +69,14 @@ class FirefoxDelegate {
   async onStopIteration() {}
 
   async beforeEachURL(runner) {
+    if (this.android && this.options.androidPower) {
+      log.info('Resetting power information');
```
soulgalore, line 73: I love info messages, that makes it so much easier to follow IRL :D

soulgalore:
Wow cool @gmierz ! This will only works on rooted phones right? Do you have an example how to turn off the power on Moto G5 so I can try it out?  I can add that to the documentation later.

soulgalore:
I think you can keep parsePowerMetric as it is now, and then later we can maybe move it out of that file, but since its Android specific I think its ok, or how do you think?

lib/firefox/webdriver/firefoxDelegate.js
gmierz: Good point! I'll move all of this start/stop code to that file.

lib/firefox/webdriver/firefoxDelegate.js
gmierz: Screen settings shouldn't affect technical/visual performance numbers very much (if at all) so it likely won't have an impact on anything that isn't a power test. I'm happy to move this to a more general setup step though if you'd like though.

gmierz:
@soulgalore I'm not sure if you need a rooted phone for this to work - all the phones I work with are rooted so I can't test that.

For the method we use to disable charging, you would need to have a rooted phone (but there are other ways of doing this, like having adb communicate with the phone over wifi - not sure if that requires root or not). You can find our code for enabling/disabling here: https://searchfox.org/mozilla-central/source/testing/raptor/raptor/power.py#17-115

There's also other preliminary code on this here: https://github.com/mozilla/powerusage-android/blob/master/utils/adb_utils.py

For MotoG5, you would run (Pixel 2 is similar, but the path to the file is `/sys/class/power_supply/battery/input_suspend`):
```
Disable: adb shell "su -c 'echo 0 > /sys/class/power_supply/battery/charging_enabled'"
Enable: adb shell "su -c 'echo 1 > /sys/class/power_supply/battery/charging_enabled'"
```

lib/firefox/webdriver/firefoxDelegate.js
gmierz: Everything is now in the Android class.

gmierz:
@soulgalore Do you want me to add some documentation for this somewhere?

gmierz:
For the CI failure, have you seen this before? I'm not sure which of my changes could have caused this: 
```
$ node ./bin/browsertime.js -b chrome test/navigationscript/navigateAndStartInTwoSteps.js -n 1 --pageCompleteCheckInactivity --timeToSettle 1000

[2020-06-02 15:06:18] INFO: Running tests using Chrome - 1 iteration(s)

[2020-06-02 15:06:19] INFO: Navigating to url https://www.sitespeed.io iteration 1

[2020-06-02 15:06:26] INFO: Start to measure

[2020-06-02 15:06:26] ERROR: Couldn't execute script named documentHeight error:JavascriptError: javascript error: Cannot read property 'scrollHeight' of null

  (Session info: chrome=83.0.4103.61)

[2020-06-02 15:06:26] ERROR: Failed to execute user script: JavascriptError: javascript error: Cannot read property 'scrollHeight' of null

  (Session info: chrome=83.0.4103.61)

[2020-06-02 15:06:26] ERROR: Unhandled value type undefined found when adding data for pageinfo 

[2020-06-02 15:06:26] ERROR: Error running browsertime TypeError: Cannot read property 'layoutShift' of undefined

    at Object.logResultLogLine (/home/travis/build/sitespeedio/browsertime/lib/support/util.js:90:16)

    at Engine.runByScript (/home/travis/build/sitespeedio/browsertime/lib/core/engine/index.js:210:10)

    at processTicksAndRejections (internal/process/task_queues.js:97:5)

    at async run (/home/travis/build/sitespeedio/browsertime/bin/browsertime.js:50:22)

The command "node ./bin/browsertime.js -b chrome test/navigationscript/navigateAndStartInTwoSteps.js -n 1 --pageCompleteCheckInactivity --timeToSettle 1000" exited with 1.
```

lib/firefox/webdriver/firefoxDelegate.js
soulgalore: No lets keep it that way then :)

soulgalore:
If you can add a PR for docs here: https://github.com/sitespeedio/sitespeed.io/tree/master/docs/documentation/sitespeed.io/mobile-phones - but you can do that later!

soulgalore:
Don't remember that error (but the following errors I should fix). I just restarted to check again (it worked on one of the tests). 

I'm working on moving the test from Travis to GitHub, will do it for sitespeed.io first and then browsertime, it seems like that environment is much more stable + faster. 

We have special code already for Moto G5 and Pixel 2, so I think we could enable/disable charging in the code ... but lets do that later in another PR :)

gmierz:
> If you can add a PR for docs here: https://github.com/sitespeedio/sitespeed.io/tree/master/docs/documentation/sitespeed.io/mobile-phones - but you can do that later!

Great! I'll get that done soon.

> Don't remember that error (but the following errors I should fix). I just restarted to check again (it worked on one of the tests).
> 
> I'm working on moving the test from Travis to GitHub, will do it for sitespeed.io first and then browsertime, it seems like that environment is much more stable + faster.

Awesome! :D

> We have special code already for Moto G5 and Pixel 2, so I think we could enable/disable charging in the code ... but lets do that later in another PR :)

Oh I didn't notice that! Sounds good to me.

soulgalore:
Thank you @gmierz :)

PR #1321, opened 2020-07-11
Make sure we collect the HAR data ASAP before we start to do other th…

soulgalore:
…ings on the page.

See https://github.com/sitespeedio/sitespeed.io/issues/3081


soulgalore:
This was fixed by another PR

PR #1332, opened 2020-08-08
Decrease the color limit when comparing if an image is full color or …

soulgalore:
…not.

We had a really high limit, trying to get of the color tearing in Docker for Chrome.
But that caused partly orange web sites to fail see:
https://github.com/sitespeedio/sitespeed.io/issues/3093#issuecomment-670683010

I think we should decrease it even more but first lets see if we can fix the
original Chrome issue again.


soulgalore:
I'm gonna rewrite visual metrics this week so I'm gonna close this one.

PR #1343, opened 2020-08-27
Check if pathname contains a path before splitting it.

gmierz:
This patch fixes an issue in `pathToFolder` that occurs when `parsedUrl.pathname` is empty. It falls back to using the given `url` as the first entry in the `pathSegments` variable when this error occurs.

I hit this issue when trying to make a measurement on `about:blank` since this url doesn't have a non-null pathname when parsed through the url module.

Peter, what do you think about this issue and solution?

### DIFF
--- a/lib/support/pathToFolder.js
+++ b/lib/support/pathToFolder.js
@@ -9,8 +9,12 @@ module.exports = function(url, options) {
     return '';
   } else {
     const useHash = options.useHash;
-    const parsedUrl = urlParser.parse(decodeURIComponent(url)),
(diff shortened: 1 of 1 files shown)


soulgalore:
Hi @gmierz does it work to measure about:blank? Or is it possible? :)

We should fix it anyway, let me test it later today.

soulgalore:
This should be safe to do so lets go!

gmierz:
Yup it works with about:blank for me when I have this patch applied. Here's a patch on our side which makes use of it for power usage testing: https://phabricator.services.mozilla.com/D88491

Issue #1344, opened 2020-08-27
Error when running commands.android.shell function: `TypeError: AndroidInstance is not a function`


gmierz:
The `commands.android.shell` command that is mentioned here doesn't seem to work for me: https://www.sitespeed.io/documentation/sitespeed.io/scripting/#add-your-own-metrics

I get the following error:
```
18:27:44    ERROR -  Exception: Could not run shell command dumpsys deviceidle whitelist +org.mozilla.geckoview_example TypeError: AndroidInstance is not a function
```

gmierz:
Thanks for the review and merge @soulgalore!

PR #1345, opened 2020-08-27
Issue #1344 - Import and use Android instead of AndroidInstance in commands.android

gmierz:
This patch fixes the issue mentioned in #1344. The problem was that `AndroidInstance` was being imported from `lib/android/` but it doesn't exist there. This patch changes the import to use `Android`, changes the class name in the commands file to AndroidCommand, and makes an additional fix (adding `new`) to get it to work.


soulgalore:
Thank you @gmierz , must have missed this when I refactored.

PR #1347, opened 2020-08-28
Fix a bug which firstPaint may be missing

sefeng211:
Our current code doesn't check `timeToNonBlankPaint` if there's
at least one paint entry. This is good if paint entries always
contain a `first-paint` entry. However, this is not the case.
Firefox only ships with first-contentful-paint.

Fix this bug by checking `timeToNonBlankPaint` even if there are
paint entries.


soulgalore:
Ah great, do you know when first-contentful-paint will be in stable?

sefeng211:
Thanks, we are aiming at 82!

Issue #1331, opened 2020-08-08
Firefox 79.0.2 on Android fails to start


soulgalore:
Started to test the newly released Firefox on Android, but it fails to start:

```
bin/browsertime.js https://www.sitespeed.io -n 1 -b firefox  --android
[2020-08-08 19:27:48] INFO: Run tests on Moto G (5) using Android version 8.1.0
[2020-08-08 19:27:48] INFO: Running tests using Firefox - 1 iteration(s)
[2020-08-08 19:27:48] INFO: Skip install HAR trigger on Android
[2020-08-08 19:28:48] INFO: Browser failed to start, trying 2 more time(s): Failed to start browser in 60 seconds.
[2020-08-08 19:28:48] INFO: Skip install HAR trigger on Android
[2020-08-08 19:29:48] INFO: Browser failed to start, trying 1 more time(s): Failed to start browser in 60 seconds.
[2020-08-08 19:29:48] INFO: Skip install HAR trigger on Android
[2020-08-08 19:30:48] INFO: Browser failed to start, trying 0 more time(s): Failed to start browser in 60 seconds.
[2020-08-08 19:30:48] ERROR: BrowserError: Could not start the browser with 3 tries
    at SeleniumRunner.start (/Users/peter/git/browsertime/lib/core/seleniumRunner.js:86:13)
```

However on the phone, the browser opens up, but the URL isn't loaded:

<img width="531" alt="Screenshot 2020-08-08 at 19 35 42" src="https://user-images.githubusercontent.com/540757/89716571-5e6d9c00-d9ae-11ea-8296-4d7820612482.png">

Maybe its some configuration that needs to be changed?

gmierz:
Hi Peter, I think this has to do with the intents and/or activity that is being used by default in browsertime. For FF79 (Fenix), we use the following command to start the app:
`[task 2020-08-07T20:59:55.894Z] 20:59:26     INFO -  adb shell_output: adb -s FA79H1A00466 wait-for-device shell su -c "am start -W -n org.mozilla.fenix/org.mozilla.fenix.IntentReceiverActivity -a android.intent.action.VIEW -d about:blank", timeout: None, timedout: None, exitcode: 0, output: Starting: Intent { act=android.intent.action.VIEW dat=about:blank cmp=org.mozilla.fenix/.IntentReceiverActivity }`

Here's the browsertime command we are using for this browser version:
[...]

PR #1379, opened 2020-09-14
Set better default for new Firefox on Android

soulgalore:
#1331

### DIFF
--- a/lib/firefox/webdriver/builder.js
+++ b/lib/firefox/webdriver/builder.js
@@ -218,11 +218,20 @@ module.exports.configureBuilder = function(builder, baseDir, options) {
       }
     } else {
       ffOptions.firefoxOptions_().androidPackage = 'org.mozilla.firefox';
-      ffOptions.firefoxOptions_().androidActivity = '.App';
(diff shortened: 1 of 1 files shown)


lib/firefox/webdriver/builder.js
```diff
@@ -218,11 +218,20 @@ module.exports.configureBuilder = function(builder, baseDir, options) {
       }
     } else {
       ffOptions.firefoxOptions_().androidPackage = 'org.mozilla.firefox';
```
gmierz, line 220: Are you leaving it as this (rather than org.mozilla.firefox_beta) in preparation for the release version?

gmierz (approved):
Looks good thanks! Just a small question about the package name.

lib/firefox/webdriver/builder.js
soulgalore: Yep, thinking there is only a couple of days more to wait?

gmierz: Sounds good. The last I heard was that it should make it into a release on around ~Sept 22nd which is about a week away.

PR #1382, opened 2020-09-18
Fix how MOZ_LOG collection works and allow custom MOZ_LOG settings.

gmierz:
This patch fixes how MOZ_LOG collection works to handle multiple `moz_log.txt*` being created. It also adds a new option so that users can set MOZ_LOG to something custom.


soulgalore:
Ah cool, thank you @gmierz !

Issue #1391, opened 2020-09-28
Profiles from the gecko profiler are incomplete and do not show the entire page load


dpalmeiro:
Any profile using --firefox.geckoProfiler shows incomplete data, for example:  https://share.firefox.dev/2S3crfW

This appears to be a regression from https://bugzilla.mozilla.org/show_bug.cgi?id=1634816 where the API has changed to use bytes instead of entries.

PR #1392, opened 2020-09-28
Increase buffer size for gecko profiles from ~1MB to 100MB

dpalmeiro:
Fixes https://github.com/sitespeedio/browsertime/issues/1391.


dpalmeiro:
Extra changes got pulled in

PR #1394, opened 2020-09-28
Fix bufferSize to proper 100MB default size.

dpalmeiro:
Sorry @soulgalore, I misread the bug.  It will eventually be changed to take bytes, but currently takes the number of entries which are 8 bytes each.  So the real value to get 100MB should be 13107200.  Sorry about that.


soulgalore:
No problem @dpalmeiro thank you for the fix.

PR #1398, opened 2020-09-29
Make sure we remove the orange and navigate using executeScript/reque…

soulgalore:
…stFrame.


soulgalore:
Hi @acreskeyMoz so I merged your fix a little to fast, been busy at work, sorry :) 

The thing is that the normal behavior changed so that we sent the JS to remove the orange and navigate in two steps in your fix, I just reverted that to follow the same pattern as before.

Issue #1142, opened 2020-01-22
[proposal] The fate of visualmetrics.py


tarekziade:
We're currently modifying visualmetrics.py enough from its original version that I think we should make it an official part of the browsertime code and not a vendored thing anymore.

To make our life easier in testing and maintaining it, I am proposing these changes:

- [x] create a "browsertime" python package in the repository and move visualmetrics.py and test_visualmetrics.py in it
- [x] make visualmetrics.py an installable script
- [x] make it a project by adding a pyproject.toml file
- [x] add Tox support so we can run lints and tests on Python 2.7 and 3.8 (with coverage)
- [x] format it using Black
- [x] add coverage information

In follow up work:

- [ ] add a version (right now it's "0.1")
- [ ] add tests until we reach 80% coverage of the module, and make it Py 3 compatible along the way

The impact for people using visualmetrics.py will be:
- it would move from vendor/ to browsertime/ directory in the repo
- it will become an installable script ("pip install browsertime")
- maybe it would make sense to rename it to avoid any confusion in the future

soulgalore:
Sounds great @tarekziade ! Lets just iterate over the naming :)

tarekziade:
work in progress in https://github.com/sitespeedio/browsertime/tree/tox

ncalexan:
I am 100% in favour of improving the existing Python script.  I'd very much like it to do its work in parallel, for example, and potentially not invoke IM to do what -- I think -- are standard image transformations.

soulgalore:
@tarekziade a side note: we also use https://github.com/paulirish/speedline to calculate speed index from the Chrome devtools log (as an alternative using Visual Metrics).

I think for Firefox using the internalWindowrecorder we could rework Speedline to calculate the same thing but for Firefox (or plain images with timestamps), that way getting Visual Metrics for Firefox would work without python/imagemagick and ffmpeg?

tarekziade:
That could make deployments easier indeed - does this tool covers the same metrics ?

soulgalore:
We will miss out Contentful Speed Index. It support First Visual Change, Speed Index, Perceptual Speed Index and Last Visual Change. That lib was done by https://github.com/pmdartus a long time ago and then later adopted by the Chrome team and used by Lighthouse. I'm thinking that the core functionality we need is there and we could either ask if they accept PRs to make it more generic or give credit and redo our own modified version?

soulgalore:
[...]

PR #1399, opened 2020-09-30
Add option for specifying logging format in visualmetrics.py.

gmierz:
This patch adds an option to specify the log formatting in the visualmetrics.py script.

### DIFF
--- a/browsertime/visualmetrics.py
+++ b/browsertime/visualmetrics.py
@@ -1975,6 +1975,11 @@ def main():
     parser.add_argument(
         "--logfile", help="Write log messages to given file instead of stdout"
     )
+    parser.add_argument(
+        '--logformat',
+        help="Formatting for the log messages",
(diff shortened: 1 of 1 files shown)


soulgalore:
Thank you @gmierz ! One thing: Do you use Python 2 or 3 in your setup? I'm wondering about https://github.com/sitespeedio/browsertime/issues/1142#issuecomment-697318411 that for me no log for Visual Metrics is created when I run Python 3.

gmierz:
Hmm, we run it with Python 3 actually and I'm able to get logs from it but we don't use log files though so I haven't checked that.

gmierz:
I checked locally with the --logfile option just now and I'm seeing logging going into the file and it being created so it sounds like something within browsertime that is causing issues.

soulgalore:
Ok cool thanks.

Issue #1419, opened 2020-10-23
Videos that are recorded after the first page iteration don't exist in the results in v10


gmierz:
After upgrading from v9 to v10, we're seeing issues with the videos from the second iteration of a pageload test: https://searchfox.org/mozilla-central/source/testing/raptor/browsertime/browsertime_pageload.js#16

We run 2 iterations with that script, and the videos from the second iteration are not available in the results even though we do see video path entries in the browsertime.json: https://treeherder.mozilla.org/#/jobs?repo=try&tier=1%2C2%2C3&revision=332da077e9b9285220feffd4df267e8ea2d9f0e2

Here's a sample archive (containing the affected browsertime.json) from our tests: https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/XcOXalPzTG2d9LsDd8tEYw/runs/0/artifacts/public/test_info/browsertime-results.tgz

In version 9 everything works as expected: https://treeherder.mozilla.org/#/jobs?repo=try&tier=1%2C2%2C3&revision=72bf66338a2b736a2902c070139fe06a282bd4b6

@soulgalore is this a bug or is there maybe a new option we need to use for this now?

soulgalore:
Hi @gmierz sorry for the late answer, this seems like a bug, nothing "should" have changed how that is handled. 

I don't fully understand the script. What's page_cycles? Do you run more than one? Then the name of the video collide when I test. Is that what happens or do you run one page_cycle and two iterations? Is there a easy way for me to reproduce it? I I tried with some simple tests but couldn't see the issue but I maybe simplified it too much.

gmierz:
That's right, in our tests we start the browser, navigate to the test page and measure, then without closing the browser, we navigate to the page again (2 page cycles). So the behavior that was there before was that the first page cycle measurement would give us data in a `query-XXXXXX` folder, then the second page cycle iteration would have its data in another `query-XXXXXX` folder. In v10 though, only the first folder is created and the second one never is (although mentions of it exist in the browsertime.json file).

Here's the command we are using: 
[...]

PR #1421, opened 2020-10-26
Navigate to the same URL should generate multiple tests.

soulgalore:
In 10 testing the same page twice in a script was broken as reported
in #1419 and also reported in https://github.com/sitespeedio/sitespeed.io/issues/3164.

The problem was that when we stop the video, we didn't know the correct URL,
that is fixed with this commit (+code cleanup to make it easier to follow).


soulgalore:
Unrelated OS X issues, so lets merge.

PR #1423, opened 2020-10-27
Remove Firefox preferences that disables the speculative connection pool

### LINKED ISSUE #1422: Firefox preferences disable the speculative connection pool
The Firefox preferences disable the network's speculative connection pool:
https://github.com/sitespeedio/browsertime/blob/325d5389b4deb6b68bff0e5b8fa42c9d457bfdc5/lib/firefox/settings/firefoxPreferences.js#L139

Speculative connections are a critical component of page load performance and thus this preference should be removed.

The history behind disabling this pref is that it was initially migrated from the Talos suite in mozilla-central.

acreskeyMoz:
This is a preference inherited from a legacy testing suite.
We do want the speculative connection pool active as it plays a critical role in pageload performance.

Closes #1422.

### DIFF
--- a/lib/firefox/settings/firefoxPreferences.js
+++ b/lib/firefox/settings/firefoxPreferences.js
@@ -135,8 +135,6 @@ module.exports = {
   'media.navigator.enabled': true,
   'media.navigator.permission.disabled': true,
   'media.peerconnection.enabled': true,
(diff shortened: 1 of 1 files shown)


soulgalore:
Thank you @acreskeyMoz !

acreskeyMoz:
Thanks for the quick review @soulgalore !

PR #1424, opened 2020-10-28
Add phone id to the HAR file.

soulgalore:


soulgalore:
Still problem with Safari installations, not related to this.

PR #1432, opened 2020-11-18
Easier to run Firefox beta/nightly

soulgalore:


soulgalore:
Failure becasue GitHub Action is not running the correct Chromedriver version yet.

PR #1435, opened 2020-11-18
Record and keep the full screen (including URL bar)

soulgalore:


gmierz:
Confirming that this works on windows. Here's an archive with the recorded videos in it: https://firefox-ci-tc.services.mozilla.com/api/queue/v1/task/aADLZsEVSSCTYBwqUtlaHg/runs/0/artifacts/public/test_info/browsertime-results.tgz

soulgalore:
Thank you @gmierz for testing :)

PR #1436, opened 2020-11-19
Try out Edge in the Docker container

soulgalore:
Hey @beenanner I've added Edge to the default browser Docker container and can do the same for sitespeed.io but I'm not sure. The size of the image will increase. We have the slim one https://github.com/sitespeedio/sitespeed.io/blob/main/Dockerfile-slim only using Firefox, but it can be that you can hit a limit on cloud instances with smaller disk size. But it was also some time ago I trimmed the images, so maybe we should look into that too before the next release.


soulgalore:
Hmm, the MSEdgeDriver is 343 MB on Linux that seems that something is wrong.

soulgalore:
Screen Shot 2020-11-20 at 8 46 10 AM

https://msedgewebdriverstorage.z22.web.core.windows.net/?prefix=88.0.702.0/

soulgalore:
Let me create a new container for Edge until they fix it.

beenanner:
Hmm do you have the Dockerfile I can tinker with in a branch @soulgalore?

soulgalore:
@beenanner https://github.com/sitespeedio/browsertime/pull/1436/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557 has Edge and then it installs Edgedriver.

soulgalore:
This has been fixed in WebDriver:
Screen Shot 2020-12-16 at 9 34 48 PM

soulgalore:
Closing see https://github.com/sitespeedio/browsertime/pull/1458

PR #1440, opened 2020-11-26
Add timeout for getting Geckoprofile data.

soulgalore:
On one of the phones where I run tests I've seen that getting the
Geckoprofile data can freeze the tests forever.


soulgalore:
Unrelated test failure.

PR #1444, opened 2020-11-30
Close system not responding popup on Android

soulgalore:


soulgalore:
Installation on Linux/OS X  broken again for GitHub actions :(

PR #1457, opened 2020-12-16
Use firefoxOptions for env settings on Android.

gmierz:
This patch fixes how environment variables are set on Android. Instead of using the `process.env` we use the `firefoxOptions.env`.


soulgalore:
Cool, thank you @gmierz !

gmierz:
Thanks for the quick merge @soulgalore :)

PR #1462, opened 2021-01-07
Hide the pretest power button click behind an option.

gmierz:
This patch hides the android power button click behind a flag since it might be causing the screen to get turned off on some phone models (Moto G5). I'm running some tests here to see: https://treeherder.mozilla.org/jobs?repo=try&revision=1cc159b1660fbe658d768169fba2a87d954adb46

You can see the bug we have about this here: https://bugzilla.mozilla.org/show_bug.cgi?id=1680998
The screen recordings are black, and the android logs also mention that the phone was going to sleep. This started after we updated to version 11.

### DIFF
--- a/lib/core/engine/index.js
+++ b/lib/core/engine/index.js
@@ -85,7 +85,9 @@ class Engine {
         throw Error(message);
       }
 
-      await android.clickPowerButton();
+      if (this.options.androidPretestPowerClick) {
+        await android.clickPowerButton();
+      }
 
       if (this.options.androidVerifyNetwork) {
         const pingAddress = get(this.options, 'androidPingAddress', '8.8.8.8');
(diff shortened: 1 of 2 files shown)


soulgalore:
Hi @gmierz thanks for creating the fix! It's strange though, where we run our tests on Moto G4 and G5 I've got different behaviour, we had the black screen problem and this fixed it.

soulgalore:
@gmierz I wonder if this is different depending on the Android version? I tested with Android 7 on my Moto G5 and there it works fine. Or do you know if you have something special in your environment? Let me test on Android 8 later today.

gmierz:
> @gmierz I wonder if this is different depending on the Android version? I tested with Android 7 on my Moto G5 and there it works fine. Or do you know if you have something special in your environment? Let me test on Android 8 later today.

:aerickson mentioned this in Riot:
```
bitbar says that setting "stay awake" in developer settings is enough to get the screen to stay on all the time if plugged in.
```

PR #1464, opened 2021-01-13
Fix backfilling fullyLoaded.

soulgalore:
If you use an alias and the URL change between runs, adding the
fullyLoaded metric breaks, because it tries to add stats to a
URL that do not exist. If the HAR file know the alias, we can
fix it. This is hacky and we should make a task to cleanup the
use of alias.


soulgalore:
There's better ways to do this.

PR #1466, opened 2021-01-15
Save gecko profile locations in the browsertime results.

gmierz:
This patch is for storing the gecko profile paths in the browsertime results. This makes it easier to figure out which gecko profiles go with which tests.


soulgalore:
Cool thank you @gmierz ! Unrelated failure, It looks like the env has changed :(

Issue #1465, opened 2021-01-14
Different URLs with the same alias do not work


soulgalore:
There's users that tests a couple of pages in a script, give each URL each an alias and then between runs, the URL will change. It could be session id or something that is in the path.

Easiest to reproduce the problem is this hack:
```
module.exports = async function(context, commands) {
  let url = 'https://www.sitespeed.io';
  if (context.index === 2) {
    url = 'https://www.sitespeed.io/documentation/';
  }
  return commands.measure.start(url, 'test');
};
```

And then run two runs. The error will look like:
```
[2021-01-14 07:29:41] ERROR: Error running browsertime TypeError: Cannot read property 'addDeep' of undefined
    at Collector.addFullyLoaded (/Users/peter/git/browsertime/lib/core/engine/collector.js:87:18)
```
The problem at that point is that we try to add the fully loaded metric (taking from the HAR) but there in the code we don't know the alias so URLs do not match. Another problem is that we for some code only have a 1-1 mapping between alias and url.

PR #1467, opened 2021-01-18
Catch if URLs change between runs.

soulgalore:
https://github.com/sitespeedio/browsertime/issues/1465


soulgalore:
Unrelated failure in Windows and OS X.

PR #1493, opened 2021-02-17
Add CLI argument to control ffmpeg thread count

TheJokr:
Hi! I'm running browsertime dockerized with a limited number of CPUs assigned to each container, but ffmpeg's default thread count seems to be based on the number of CPUs of the entire system. This PR adds a new `videoParams.threads` CLI argument to overwrite ffmpeg's thread count for this scenario. The behavior will be the same if the option is not specified, i.e., ffmpeg picks the number of threads.


soulgalore:
Thank you @TheJokr !

PR #1495, opened 2021-02-25
Add perf stats support to browsertime

dpalmeiro:
Firefox has an infrastructure to provide internal micro timings and emit the output as a JSON.  We'd like to extend browsertime to automatically measure and collect this data when requested.


soulgalore:
Cool thank you @dpalmeiro I'll add a test case for one of the GitHub actions to test it.

PR #1498, opened 2021-03-05
Move Firefox perfstats results to browsertime.json

dpalmeiro:
Since the perfstats data is not large and it is mostly internal metric measurements, it seems a better fit to output the results in browsertime.json instead of separate files.


lib/core/engine/collector.js
```diff
@@ -357,6 +357,14 @@ class Collector {
         });
       }
 
+      // Add Firefox perfStats if available
+      if (this.options.firefox && this.options.firefox.perfStats) {
+        if (!results.perfStats) {
+          results.perfStats = [];
```
soulgalore, line 363: Hi @dpalmeiro sorry I've been offline for a week, I'm back now. I haven't tested the perfStats yet, but what about just rename the field _perfStats_ to something more descriptive? Is it internal browser performance statistics or what is collected?

soulgalore:
Also @dpalmeiro do you have a link to documentation or a task about the flags in `--firefox.perfStatsParams.mask`?

dpalmeiro:
@soulgalore The current flags in the mask are here:
https://searchfox.org/mozilla-central/source/tools/performance/PerfStats.h#24-33

lib/core/engine/collector.js
dpalmeiro: Sure, I'm open to any naming changes.  Maybe geckoPerfStats or something if that is more clear, or anything you may think sounds better.

soulgalore: Yes, geckoPerfStats sounds get better, I'm just thinking so it do not collide with something else in the future. I can change it if you want, I wanna push the release later tonight.

dpalmeiro: Sure, please go ahead. Thanks!

Issue #1494, opened 2021-02-24
Please honour the *_proxy environment variables


paddy-hack:
I guess I got spoiled by the many utilities that support `https_proxy`, `http_proxy` and `no_proxy` environment variables (or their uppercase cousins) out of the box.  I was surprised `sitespeed.io` doesn't and requires me to add command-line options to set the proxies.  These command-line option do not work when passed the typical values for the environment variables so these need some `sed` processing before they can be used as option values.  Moreover, support for `no_proxy` seems to be missing altogether, making it impossible to exempt certain locations from going through a proxy.

Searching the project here on `http_proxy` or `https_proxy` turns up several examples where people take the trouble to pass those environment variables to their Docker container, either via `docker`'s `-e` option or a Docker Compose file's `environment` attribute.  It would be so much easier for everyone if `sitespeed.io` honours these environment variables (and uses command-line arguments to override, preserving current behaviour).

I would prefer support for the lower case variant of the environment variables as those seem to be most widely supported in my experience (and [this blog post](https://about.gitlab.com/blog/2021/01/27/we-need-to-talk-no-proxy/)).  If you want to support the uppercase variants as well, I suggest uppercase → lowercase → command-line argument order in terms of increasing precedence.  

For `no_proxy`, please make sure to support CIDR notation because adding `10.0.0.0/8` one IP address at a time is no fun :wink:

soulgalore:
Hi @paddy-hack I usually recommend people to test without a proxy to make sure it doesn't interfere with any metric, but if the implementation is simple and there are use cases for using a proxy,  I can review a PR :) Right now we use the default in WebDriver/Selenium. I moved the issue to Browsertime since it's cleaner if the implementation happens there.

paddy-hack:
I can understand the desire to cut out factors you cannot control, like proxies, when doing development.  Yet, it is also important to know how one's *production* site is doing in terms of performance.

My issue arose when running tests against a(n example) production site in CI/CD, where the CI/CD runners/workers were behind (on the other side of) a proxy.  I have been dealing with proxies a fair deal and have gotten much used to things working out-of-the box if the `https_proxy`, `http_proxy` and `no_proxy` settings are correct.  For `sitespeed.io` that didn't work.

As for a PR, I'll leave that for someone with more free time on their hands.  In the mean time, I can work-around with a bit of `sed` magic and those "awkward" proxy-related command-line options you provide.  At least, until the need for `no_proxy` pops up its head ...

[...]

Issue #857, opened 2019-06-11
About Proxy vs NoProxy 


softwareklinic:
Is it possible to use NO PROXY for just the PAGE itself -- but all the resources on page are still going thru PROXY -- while testing on AWS?

Any thoughts..?

soulgalore:
Hi @softwareklinic no I don't think so. Whats your use case, what do you need to do?

Best
Peter

soulgalore:
Closing, let me know if you still need it and please give me the use case :)

Issue #609, opened 2018-06-14
proxy.pac usage in browsertime or docker run


vinay100sg:
Hi @soulgalore ,

Could you please advise if there is a way to use a proxy.pac file in the docker run -browsertime options.

--browsertime.proxy.https [=host:port] works fine when we have all the domains resolved by the page hit is behind the proxy. The issue is, in our test environment the main (https://**main_domain.com**/login.html) page doesn't need to go through proxy; however, some of the resources (eg: https://**sub_domain.com**/required.png) are behind the proxy as the content is directly accessed from prod to construct the complete page. So, when hit the main page without proxy set we can see only html page without resources.

Simplest way we are thinking is to use the rules defined in the proxy.pac file to address this issue. However, if we give the option something like below,

--browsertime.proxy.https=http://XXX:pppp/proxy.pac OR
--proxy=http://XXX:pppp/proxy.pac

both options seem to be ignored and the test is running without proxy. Any help to address this would be highly appreciated.

Thanks,
Vinay

soulgalore:
Hi @vinay100sg 
we are missing support for pac files but it looks like Selenium supports it see https://seleniumhq.github.io/selenium/docs/api/javascript/module/selenium-webdriver/proxy.html

We need to add an extra CLI param for the pac configuration and do some small changes when we setup the proxy. Do you have time to do a PR? We can help out to guide you?

Best
Peter

vinay100sg:
Thanks @soulgalore ! Happy to contribute. I will try my best, kindly guide through.

vinay100sg:
Hi @soulgalore ,

We made few changes in our local docker image and tried to test with a pac file. However, getting below error. 

**Error:**
[2018-06-26 06:03:30] INFO: Browser failed to start, trying one more time: unknown error: cannot parse capability: proxy
from unknown error: 'proxyAutoconfigUrl' must be a string
  (Driver info: chromedriver=2.40.565383 (76257d1ab79276b2d53ee976b2c3e3b9f335cde7),platform=Linux 3.10.0-693.21.1.el7.x86_64 x86_64)
[2018-06-26 06:03:30] INFO: Browser failed to start, trying one more time: unknown error: cannot parse capability: proxy
from unknown error: 'proxyAutoconfigUrl' must be a string
[...]

PR #1524, opened 2021-03-29
Improve proxy configuration support

paddy-hack:
This is *untested* work I did while diving into the proxy support of `sitespeed.io` before finding out that on Linux systems the "normal" environment variables are honoured after all (#1494).

The changes merely forward the command-line settings straight to the underlying `selenium-webdriver` API.  As long as the values adhere to that module's API specification, things ought to work but I have not tested this.  The changes might be useful for folks using non-Linux systems.

See also #609 and #857.


soulgalore:
Hi @paddy-hack cool, how the easiest way to test this?

paddy-hack:
I guess pretty much the same way as you tested the HTTP and HTTPS proxy configuration support.

soulgalore:
Thanks @paddy-hack for your contribution!

PR #1529, opened 2021-04-01
Display standard deviation instead of the home made median deviation

soulgalore:


soulgalore:
Safari builds are broken (again)

PR #1532, opened 2021-04-05
Fix blockDomainsExcept using WebPageReplay.

soulgalore:
See https://github.com/sitespeedio/sitespeed.io/pull/3329

And thank you https://github.com/inderpartap

### DIFF
--- a/docker/scripts/start.sh
+++ b/docker/scripts/start.sh
@@ -95,7 +95,7 @@ function runWebPageReplay() {
   wpr record $WPR_PARAMS > /tmp/wpr-record.log 2>&1 &
   record_pid=$!
   sleep $RECORD_WAIT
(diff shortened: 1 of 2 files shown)


SunnyGurnani:
Great job @inderpartap and thank you @soulgalore for guidance and merging this PR.

WebPageReplay along with 3rd Party Block Domain is a an amazing feature to test performance of the site solely on the basis of developer code.

soulgalore:
@inderpartap it seems the alias in the cli for the `--chrome.blockDomainsExcept` doesn't work or if I have missed something, if I add the browsertime prefix, it works for me:

`docker run --cap-add=NET_ADMIN --rm -v "$(pwd):/sitespeed.io" -e REPLAY=true -e LATENCY=100 sitespeedio/sitespeed.io -n 1 --firstParty ".*(move.com|realtor.com|moveaws.com|rdcpix.com).*" --browsertime.chrome.blockDomainsExcept "*.realtor.com" https://www.realtor.com/`

Can you check if that works for you too? I'll try to fix this later + add documentation about blocking.

inderpartap:
@soulgalore Thanks for getting back. I tested it out with the `browsertime` prefix, and it worked like a charm. 
The cli for `--chrome.blockDomainsExcept` works when testing without replay so maybe there would be something else that was missed.

Issue #1550, opened 2021-04-20
Fix scroll.byPages command to work in all browsers


soulgalore:
I think this should be possible in plain JS, let me test.

soulgalore:
Coming in next release.

PR #1551, opened 2021-04-21
Add generic scroll by page #1550

soulgalore:


soulgalore:
Failing since Chrome isn't updated on Windows yet (or at least in the GitHub Action.

Issue #1557, opened 2021-04-26
Script in browser are broken for all browsers except Firefox


soulgalore:
All browsers except Firefox fails to get the data in the browser scripts located in:
https://github.com/sitespeedio/browsertime/tree/main/browserscripts/browser

This has been broken since privileged scripts was introduced. Let me fix that.

PR #1558, opened 2021-04-26
Only run scripts that needs privileges in Firefox.

soulgalore:
https://github.com/sitespeedio/browsertime/issues/1557


soulgalore:
Hi @gmierz put you on review FYI but gonna merge asap :) 

This has been broken since we merged it: The appconstants script that needs privileges breaks the "browser" scripts in other browsers than Firefox (we miss it in the output.json). A couple of levels down we do a check for Firefox and then throws and error https://github.com/sitespeedio/browsertime/blob/main/lib/core/seleniumRunner.js#L390-L395 and that is swallow, not seen.

With this fix all scripts in the browser category will work in all browsers.

PR #1581, opened 2021-05-25
Update getViewPort Pixel 2 size to use x as a separator

radum:


soulgalore:
Thank you @radum for the fix!

PR #1601, opened 2021-06-24
Add firefox memory support to browsertime

dpalmeiro:
Firefox is capable of dumping a memory report for all of it's processes.  This patch adds support to browsertime to do this after each iteration and then parse the report to measure the total resident memory of the browser.  The memory report is also saved if other details are required instead.  A new option "--firefox.memoryReport" will enable this feature.


soulgalore:
Hi @dpalmeiro ah this is cool, great! I just wanna test it on Android too, I'm gonna do that this weekend.

PR #1611, opened 2021-07-19
Check for resultsDir cli option before creating default directory

ledleds:
I've noticed that the `browsertime-results` directory gets created even if the resultsDir option is passed in the cli. 

So in this PR we check if we have that option before trying to create it. 

I'm even wondering if this creation needs to happen here as in lib/support/storageManager.js it calls mkdir from fs with recursive true so from hat I can see it would try to make it here again. If this is true lines 47 to 52 of browsertime.js can just be removed I guess :)


soulgalore:
Hi @ledleds ah thanks I missed that!

PR #1620, opened 2021-07-30
cli flag to specify chromedriver port

svkrclg:
https://github.com/sitespeedio/sitespeed.io/issues/3427

### DIFF
--- a/lib/chrome/webdriver/builder.js
+++ b/lib/chrome/webdriver/builder.js
@@ -25,6 +25,9 @@ module.exports.configureBuilder = function(builder, baseDir, options) {
     }
 
     const serviceBuilder = new chrome.ServiceBuilder(chromedriverPath);
+    if (options.chrome.chromedriverPort) {
+      serviceBuilder.setPort(options.chrome.chromedriverPort);
+    }
     if (
       options.verbose >= 2 ||
       chromeConfig.enableChromeDriverLog ||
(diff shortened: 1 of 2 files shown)


svkrclg:
I am new in this org as a contributor, let me know what else needs to be added.

soulgalore:
Hi @svkrclg looks great, thanks the addition! I'm on vacation, that's why the late merge.

lib/chrome/webdriver/builder.js
```diff
@@ -25,6 +25,9 @@ module.exports.configureBuilder = function(builder, baseDir, options) {
     }
 
     const serviceBuilder = new chrome.ServiceBuilder(chromedriverPath);
+    if (options.chrome.chromedriverPort) {
```
soulgalore, line 28: This need either use lodash.get or you can add a check like:
if (options.chrome  && options.chrome.chromedriverPort) ...

The reason is that even when you run Chrome, we aren't 100% you have set any options for Chrome (that's the way it works right now at least).

soulgalore (changes requested):
One minor thing!

lib/chrome/webdriver/builder.js
svkrclg: Yes, sure thing.

svkrclg:
Also, do you think it is a good idea to check whether the port is free or not before starting chromedriver?

soulgalore:
@svkrclg no I think that could up to the developer that starts the driver just as long as we push an error to the log if it fails.

svkrclg:
In case of an invalid port, the Chromedriver error is pushed to `chromedriver.log `if verbose >=2

soulgalore:
Thank you @svkrclg !!!

PR #1624, opened 2021-08-16
Increase bottom margin for Visual Metrics.

soulgalore:
See https://phabricator.wikimedia.org/T288927


soulgalore:
No I've checked the margin we have today should be enough.

PR #1650, opened 2021-10-06
Add an option to prevent browsertime from setting Firefox default preferences.

gmierz:
This patch adds an option (--firefox.noPrefs) that lets us prevent browsertime from setting default preferences. This solves an issue where settings in an existing profile get overriden by the preferences set in browsertime.

### DIFF
--- a/lib/firefox/webdriver/builder.js
+++ b/lib/firefox/webdriver/builder.js
@@ -83,9 +83,11 @@ module.exports.configureBuilder = function(builder, baseDir, options) {
     }
   }
 
-  Object.keys(defaultFirefoxPreferences).forEach(function(pref) {
-    ffOptions.setPreference(pref, defaultFirefoxPreferences[pref]);
-  });
+  if (!firefoxConfig.noPrefs) {
+    Object.keys(defaultFirefoxPreferences).forEach(function(pref) {
+      ffOptions.setPreference(pref, defaultFirefoxPreferences[pref]);
+    });
+  }
 
   if (firefoxConfig.disableSafeBrowsing) {
     Object.keys(disableSafeBrowsingPreferences).forEach(function(pref) {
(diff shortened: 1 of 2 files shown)


gmierz:
Hi @soulgalore! We have an issue where our prefs are being overriden by the settings in browsertime - this patch would let us prevent that.

soulgalore:
Hi @gmierz aha cool, what about changing the name to `--firefox.noDefaultPrefs`, thinking its a little more self explaining?

lib/support/cli.js
```diff
@@ -483,6 +483,12 @@ module.exports.parseCommandLine = function parseCommandLine() {
       type: 'boolean',
       group: 'firefox'
     })
+    .option('firefox.noPrefs', {
```
soulgalore, line 486: Thinking noDefaultPrefs maybe is a little better?

lib/firefox/webdriver/builder.js
```diff
@@ -83,9 +83,13 @@ module.exports.configureBuilder = function(builder, baseDir, options) {
...
 
-  Object.keys(defaultFirefoxPreferences).forEach(function(pref) {
-    ffOptions.setPreference(pref, defaultFirefoxPreferences[pref]);
-  });
+  if (!firefoxConfig.noDefaultPrefs) {
+    Object.keys(defaultFirefoxPreferences).forEach(function(pref) {
+      ffOptions.setPreference(pref, defaultFirefoxPreferences[pref]);
+    });
+  } else {
+    log.info('Skip setting default preferences for Firefox');
```
gmierz, line 91: I saw that you added a log in the chrome PR so I've added one here as well for parity.

soulgalore: Great!

soulgalore (approved):

soulgalore:
Thank you @gmierz :)

Issue #1662, opened 2021-10-20
Firefox geckoprofiles and memory reports not collected on Android 10, 11 unrooted devices


acreskeyMoz:
Due to the introduction of Scoped Storage in Android 10, applications can no longer directly access the path `/sdcard` on unrooted devices:

This breaks the Firefox features of writing geckoprofiles and memory reports on Android
https://github.com/sitespeedio/browsertime/blob/bcdaeaebbbaa40c845e6171526c642668852c573/lib/firefox/geckoProfiler.js#L106

https://github.com/sitespeedio/browsertime/blob/7b0ba3106563490c041616e202f0f4c4f9c0aef3/lib/firefox/memoryReport.js#L79

I'm not an expert in the android file system, but writing to the package name within `sdcard` is permitted.
e.g.

writing to, for instance, ```/sdcard/Android/data/org.mozilla.firefox_beta/files/geckoProfile-${index}.json```

PR #1663, opened 2021-10-20
Fix to firefox for android profile and memory report serialization

acreskeyMoz:
On unrooted devices running Android 10 and later, we do no longer have access to the root of `/sdcard`

https://github.com/sitespeedio/browsertime/issues/1662


soulgalore:
Hi @acreskeyMoz aha thanks. I don't have a phone with Android 10 but I can test on a couple of other devices just to verify. If you can run ` npm run lint:fix` it will fix the linting issues.

PR #1668, opened 2021-11-08
Loop until we find a frame with a good viewport or until we run out of retries.

gmierz:
This patch fixes an issue that occurs when the first frame in the video is not a solid color (e.g. a webpage is shown). In those cases, the viewport width and height end up as 0 because the colors are not similar (between the middle and edge).

With this patch, we run through up to the first 5 (by default) frames until we find a good viewport. If none can be found, then we log an exception. The maximum number of frames to check can be modified with `--viewportretries X`. Setting `X` to 1 will give the same behaviour as we had prior to this patch.

### DIFF
--- a/browsertime/visualmetrics.py
+++ b/browsertime/visualmetrics.py
@@ -44,12 +44,12 @@ import shutil
 import subprocess
 import tempfile
 
-if (sys.version_info > (3, 0)):
-    GZIP_TEXT = 'wt'
-    GZIP_READ_TEXT = 'rt'
+if sys.version_info > (3, 0):
+    GZIP_TEXT = "wt"
+    GZIP_READ_TEXT = "rt"
 else:
-    GZIP_TEXT = 'w'
-    GZIP_READ_TEXT = 'r'
+    GZIP_TEXT = "w"
+    GZIP_READ_TEXT = "r"
 
 # Globals
 options = None
@@ -178,10 +178,10 @@ def extract_frames(video, directory, full_resolution, viewport):
         ]
         logging.debug(" ".join(command))
         lines = []
-        if (sys.version_info > (3, 0)):
-             proc = subprocess.Popen(command, stderr=subprocess.PIPE, encoding='UTF-8')
+        if sys.version_info > (3, 0):
+            proc = subprocess.Popen(command, stderr=subprocess.PIPE, encoding="UTF-8")
         else:
-             proc = subprocess.Popen(command, stderr=subprocess.PIPE)
+            proc = subprocess.Popen(command, stderr=subprocess.PIPE)
         while proc.poll() is None:
             lines.extend(iter(proc.stderr.readline, ""))
 
@@ -370,7 +370,7 @@ def find_video_viewport(
     viewport_time,
     viewport_retries,
     viewport_min_height,
-    viewport_min_width
+    viewport_min_width,
 ):
     logging.debug("Finding Video Viewport...")
     viewport = None
@@ -380,9 +380,9 @@ def find_video_viewport(
         retries = -1
 
         while (
-            viewport is None or
-            viewport["height"] <= viewport_min_height or
-            viewport["width"] <= viewport_min_width
+            viewport is None
+            or viewport["height"] <= viewport_min_height
+            or viewport["width"] <= viewport_min_width
         ):
             retries += 1
             if retries >= 1:
@@ -392,7 +392,8 @@ def find_video_viewport(
                 # of frames to check.
                 if retries >= viewport_retries:
                     logging.exception(
-                        "Could not calculate a viewport after %s tries.", viewport_retries
+                        "Could not calculate a viewport after %s tries.",
+                        viewport_retries,
                     )
                     break
                 logging.info("Failed to find a good viewport. Retrying...")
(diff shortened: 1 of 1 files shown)


gmierz:
https://user-images.githubusercontent.com/10966989/140789791-b6b3a9ee-8ccc-48e0-bc72-855d6602d740.mp4

Here's a sample video that the existing visualmetrics.py script fails on.

soulgalore:
Hi @gmierz ah great! :)

gmierz:
I still have an issue with these changes though that happens when the width/height are close to 0 but not equal to 0. I'm not sure how this should be taken care of - I think it's clear that we'll never want to crop a viewport of 1x5 but how could we make this decision in the script.

One idea I have is to use half the image width/height as a trigger a retry but in the case where we have a viewport occupying less than half the screen, this would fail.

@soulgalore what do you think if I change the `0` to `50` instead? That value seems to work fine for me with my failing videos.

gmierz:
Actually, what do you think about me adding a command line flag to change this value from 0 to 50 or another value?

soulgalore:
@gmierz sounds great add the flag for now.

soulgalore:
@gmierz is there an easy way of for me to replicate the issue so I can try before/after the fix? I'm thinking this do not happen in every case right?

Is this the use case?
```
module.exports = async function(context, commands) {
  await commands.navigate('https://www.sitespeed.io');
  return commands.measure.start('https://www.sitespeed.io/documentation/');
};
```

gmierz:
Here's the test that this happened on very often but still intermittently. You just need to add the `commands.` in front of each of the steps:
```
    ["measure.start", "billg"],
    ["navigate", "https://www.reddit.com/user/thisisbillgates/"],
    ["wait.byTime", 500],
    ["measure.stop", ""],
    ["click.byXpathAndWait", "/html/body/div[1]/div/div[2]/div[2]/div/div/div/div[2]/div[2]/div/div/div/a[3]"],
    ["wait.byTime", 500],
    ["measure.start", "hot"],
    ["click.byXpathAndWait", "/html/body/div[1]/div/div[2]/div[2]/div/div/div/div[2]/div[4]/div[1]/div[1]/div[2]/a[2]"],
    ["wait.byTime", 500],
    ["measure.stop", ""],
    ["measure.start", "top"],
    ["click.byXpathAndWait", "/html/body/div[1]/div/div[2]/div[2]/div/div/div/div[2]/div[4]/div[1]/div[1]/div[2]/a[3]"],
    ["wait.byTime", 500],
    ["measure.stop", ""],
    ["wait.byTime", 500],
```

It's the `hot` and `top` tests which fail most often. Here's a try run with the failures: https://treeherder.mozilla.org/jobs?repo=try&selectedTaskRun=dqfXF8DuTgeWGwsk4rD1Wg.0&revision=698a2070a92264a0b41c16e3f4b5d4f9142ec0e1

Here's a try run with the current patch applied (using 100 as the min height/width): https://treeherder.mozilla.org/jobs?repo=try&selectedTaskRun=dqfXF8DuTgeWGwsk4rD1Wg.0&revision=37a8578a220edbdeb69f74fa2928d3e3544eaa67

gmierz:
@soulgalore Also, I should mention that I've only seen this happen on Reddit when navigating so far. There may be other tests that we have the failure on, but they occur much less frequently.

soulgalore:
@gmierz feel free to merge this if it makes it work for you. I had a look at the script but that last step doesn't work on the live site anymore (I think). I will test some more later tonight.

In your scripting, I would try to get rid of of those wait for 500 ms by changing when to end the tests. Default is to wait for the onload to happen + 2 seconds and if you change that to `--pageCompleteCheckInactivity --pageCompleteWaitTime 5000` that makes it work for me (=wait for 5 seconds on inactivity in the Resource Timing API).

gmierz:
@soulgalore the patch will definitely fix some intermittent issues for us in other tests so I'll merge it but before I do, I'll test out your suggestions to see if that fixes it.

gmierz:
With those changes, the Linux tests still fail because of this issue. Here's a sample video:

https://user-images.githubusercontent.com/10966989/141191106-2d54d011-9813-42b8-ae86-2d2f2e74f7a9.mp4

gmierz:
> @gmierz feel free to merge this if it makes it work for you. I had a look at the script but that last step doesn't work on the live site anymore (I think). I will test some more later tonight.
> 

I forgot to mention that we've been using recorded sites in these tests.

soulgalore:
@gmierz I didn't fully understood 
> With those changes, the Linux tests still fail because of this issue. Here's a sample video:

You mean the changes I proposed or your visual metrics fix? :) The scripting changes was only to make the code cleaner.

gmierz:
Haha oh whoops! I thought you were suggesting some fixes, I'll merge this when I get back from the holiday. :)

Thanks!

soulgalore:
Sorry I wasn't clear. If you run with `--pageCompleteCheckInactivity ` for test that do Ajax requests you can probably just remove all of those  wait.byTime so the code becomes simpler :)

gmierz:
> Sorry I wasn't clear. If you run with `--pageCompleteCheckInactivity ` for test that do Ajax requests you can probably just remove all of those wait.byTime so the code becomes simpler :)

Ah wonderful! Thanks for the tips :)

PR #1680, opened 2021-11-22
Use the viewport to determine if more cropping is needed.

gmierz:
This patch adds a new variable as output from the `find_video_viewport` function that can be used to determine if the frames will be cropped or not. This is then used in the frame de-duplication, and render-start functions to prevent the top of the frame from being cropped out. Before this patch, the top of the test page was being removed and this had caused issues in tests which first showed a banner on the top of the page.

See here for conversations about this issue: https://github.com/sitespeedio/browsertime/commit/cee2030d1ae5a5d354b41fdfa0f61b53ce36b25c

### DIFF
--- a/browsertime/visualmetrics.py
+++ b/browsertime/visualmetrics.py
@@ -466,7 +466,6 @@ def find_video_viewport(
                     viewport = find_image_viewport(frame)
                 else:
                     viewport = {"x": 0, "y": 0, "width": width, "height": height}
-                    cropped = False
 
                 os.remove(frame)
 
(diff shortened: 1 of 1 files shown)


soulgalore:
Thanks @gmierz for all the work with this, this is great! I'll run some tests first thing tomorrow.

soulgalore:
Thank you @gmierz ! I did some local time testing, worked fine. I'm gonna do a release and push it on the test server later today.

soulgalore:
@gmierz had this running for a couple of hours and everything looks good, thanks again! :)

gmierz:
That's fantastic, thanks for testing and merging it :)

PR #1687, opened 2021-12-09
Remove the top 10 rows of the image to handle progress bars on some mobile browser recordings.

gmierz:
On the GeckoView browser, there's a progress bar that shows up at the top of the viewport and it gets included in the extracted frames. This causes contentfulspeedindex to regress considerably.

Here's an example of this: https://treeherder.mozilla.org/perfherder/graphs?highlightAlerts=1&highlightChangelogData=1&highlightCommonAlerts=0&selected=3757587,1449972123&series=autoland,3757587,1,13&timerange=2592000

The patch which fixed an issue with double-cropping on desktop caused an issue on mobile for GeckoView as we no longer do extra cropping if the page was already cropped. This left the progress bar at the top of the page. Here's the regressing patch: https://github.com/sitespeedio/browsertime/commit/f27fdac113108f0da72c347b5585f0e49da0f54a

Furthermore, irregardless of the changes from this patch, the viewport was being cropped far too much on the bottom (taking out about 30-50 rows which is a good portion of the image). This will be fixed with this patch. I added a method to determine if we have a mobile or desktop recording  (using ffprobe) and modify the cropping parameters based on that.

Here is an example of what the viewport looked like before the regressing patch (note how there is far too much cropping going on):
without_double_cropping_fix

Next, this is what it looked like when the double-cropping fix patch landed (note the progress bar at the top):
without_patch_cropping

Finally, this is what it looks like with this patch applied (no more progress bar, and no more losing the bottom of the page):
with_current_patch


gmierz:
Hey @soulgalore! Let me know if you have any questions about this change or if you have any ideas for a different approach to this.

I went through a couple other solutions before arriving at this one:
(1) Dynamic method to find a bar at the top of the page (using gradients). This method is flaky though because it might find a "progress bar" that isn't actually a progress bar.
(2) Change the orange frame size to exclude the progress bar. I felt that this was too strong of a move because if we don't always have a full orange frame, then in some cases, we'll never be able to extract the top of the page using the orange frame. In other words, we'll have lost information.

An alternative solution to this was adding command line options for specifying these hardcoded cropping parameters since it seems that they depend quite a bit on the user's use-case.

soulgalore:
Hi @gmierz let me run some tests tomorrow. Usually when I've done (small) changes in the past I break Chrome, when run Chrome in emulated mobile mode and they have some small changes going on in the bottom of the page, so I always try to do some tests for that.

FYI: The `is mobile` check: I've implemented video on Safari on IOS and I'm waiting on a new release of https://github.com/danielpaulus/quicktime_video_hack for that to work (so is mobile will then mean is android I guess).

> An alternative solution to this was adding command line options for specifying these hardcoded cropping parameters since it seems that they depend quite a bit on the user's use-case.

Yeah maybe in the future for "hardcore" users as your team. This seems like a hard problem to solve.

gmierz:
> > An alternative solution to this was adding command line options for specifying these hardcoded cropping parameters since it seems that they depend quite a bit on the user's use-case.
> 
> Yeah maybe in the future for "hardcore" users as your team. This seems like a hard problem to solve.

I'm happy to implement something for that now if that's ok with you. It sounds like I'd need to modify this patch to handle the iOS case anyways. Using a command-line option will also mean that I won't be making changes that are browser-specific in the processing script.

soulgalore:
@gmierz It's fine to run like this for me as long as it work for you. Command line for cropping parameters seems like it should be like the last resort? For the iPhone ... lets wait until the video work and I can make a PR for that and you can assist?

gmierz:
> @gmierz It's fine to run like this for me as long as it work for you. Command line for cropping parameters seems like it should be like the last resort? For the iPhone ... lets wait until the video work and I can make a PR for that and you can assist?

@soulgalore I agree, it works for me, and if it works for you then I'd say we're good. +1 for waiting until the iOS video work.

Does the patch look good to you as-is or is there anything you'd like me to change?

soulgalore:
@gmierz looks good, I have one question: ffprobe do you know if there are any chance its not installed when you installs ffmpeg? Thinking we adding a new requirement but maybe its always installed side by side?

gmierz:
@soulgalore I just checked and on Ubuntu ffprobe is installed alongside ffmpeg: https://packages.ubuntu.com/bionic/video/ffmpeg

When using brew, it's apparently also installed alongside ffmpeg (I can't confirm this as I don't have a mac): https://stackoverflow.com/a/35627930/4700298

soulgalore:
Great, thanks for checking!

soulgalore:
@gmierz yes please go ahead :)

gmierz:
Thanks @soulgalore!

PR #1688, opened 2021-12-30
Add support for appending to user agent for Chrome https://github.com…

soulgalore:
…/sitespeedio/sitespeed.io/issues/3118


soulgalore:
FF 95 is again broken with the HAR plugin, that's why the failures.

Issue #1671, opened 2021-11-15
Firefox HAR export trigger error for Firefox 94


soulgalore:
Some tests for Firefox on our Github Actions tests has stopped working. I started to see that on Windows and now [also on Linux](https://github.com/sitespeedio/browsertime/runs/4216597356?check_suite_focus=true):

The problem is that it generates an empty HAR file:
```json
{
"version":"1.2",
"creator":{"name":"Firefox","version":"94.0.1"},
"browser":{"name":"Firefox","version":"94.0.1"},
"pages":[],
"entries":[]
}
```

However if try the same on http://softwareishard.com/test/harexporttrigger/ I can see that it works. So something in the Browsertime setup etc makes it fail for 94. Running Browsertime locally on Mac with 94 also fails.

soulgalore:
Added 93 on the Linux Github action https://github.com/sitespeedio/browsertime/pull/1674 and then it works again.

soulgalore:
This works if I increase the timeToSettle time. Increasing it to 7 seconds make it work on my machine.

soulgalore:
But settle time do not fix everything, In one of the tests where we test multiple pages wi still get the problem:
```
 bin/browsertime.js -b firefox test/navigationscript/multi.js https://www.sitespeed.io/blog/ -n 1 
[2021-11-17 11:07:09] INFO: Running tests using Firefox - 1 iteration(s)
[2021-11-17 11:07:27] INFO: Testing url https://www.sitespeed.io iteration 1
[2021-11-17 11:07:37] INFO: Waiting on har-export-trigger to collect the HAR
[2021-11-17 11:07:37] INFO: Testing url https://www.sitespeed.io/examples/ iteration 1
[2021-11-17 11:07:46] INFO: Waiting on har-export-trigger to collect the HAR
[2021-11-17 11:07:47] INFO: Testing url https://www.sitespeed.io/documentation/ iteration 1
[2021-11-17 11:07:56] INFO: Waiting on har-export-trigger to collect the HAR
[2021-11-17 11:07:56] INFO: Testing url https://www.sitespeed.io/blog/ iteration 1
[2021-11-17 11:08:06] INFO: Waiting on har-export-trigger to collect the HAR
[...]

PR #1690, opened 2021-12-30
Use FF 94 for tests

soulgalore:


soulgalore:
See https://github.com/sitespeedio/browsertime/issues/1671#issuecomment-999412035

PR #1716, opened 2022-02-08
Remove persist log since https://github.com/sitespeedio/browsertime/i…

soulgalore:
…ssues/1671 is fixed


soulgalore:
Windows is still on Firefox 96

PR #1717, opened 2022-02-09
Always disable the ChromeWhatsNewUI splash screen

soulgalore:


soulgalore:
Windows is still on Firefox 96

PR #1741, opened 2022-03-29
Replace ImageMagick with Python dependencies.

gmierz:
This patch replaces ImageMagick with two new Python dependencies: (i) OpenCV-Python, and (ii) Numpy.

The values obtained from tests using this script produce an exact match between the old version and this verion: https://treeherder.mozilla.org/jobs?repo=try&revision=c5bf570c78a1c0533e1eb7872760ebd6a07b9e5c

In the browsertime tasks found there, those suffixed with -vismet use the old script, and those without the suffix use the new script.

One thing to note is that the CSI value will be different as the new method for edge detection differs slightly (in a good way) from the old one.

Alongside these changes, I've added a new feature hidden behind the --contentful-video flag. When set, two new videos of the edges used in the CSI calculation will be produced in the same location as the source video.

Lastly, all usages/mentions of ImageMagick have been removed from the script.

### DIFF
--- a/lib/video/postprocessing/visualmetrics/visualMetrics.js
+++ b/lib/video/postprocessing/visualmetrics/visualMetrics.js
@@ -111,6 +111,8 @@ module.exports = {
 
     scriptArgs.unshift(SCRIPT_PATH);
 
+    log.info("sparky")
+    log.info(process.env.PYTHON)
     log.debug('Running visualmetrics.py ' + scriptArgs.join(' '));
     log.info('Get visual metrics from the video');
     try {
--- a/lib/video/postprocessing/visualmetrics/visualMetrics.js
+++ b/lib/video/postprocessing/visualmetrics/visualMetrics.js
@@ -49,7 +49,7 @@ module.exports = {
       '--viewport',
       '--viewportretries',
       60,
-      '--viewportminheight'
+      '--viewportminheight',
       100,
       '--viewportminwidth',
       100,
--- a/lib/video/postprocessing/visualmetrics/visualMetrics.js
+++ b/lib/video/postprocessing/visualmetrics/visualMetrics.js
@@ -46,7 +46,13 @@ module.exports = {
       '--renderignore',
       5,
       '--json',
-      '--viewport'
+      '--viewport',
+      '--viewportretries',
+      60,
+      '--viewportminheight'
+      100,
+      '--viewportminwidth',
+      100,
     ];
 
     if (options.visualMetricsPerceptual) {
--- a/lib/video/postprocessing/visualmetrics/visualMetrics.js
+++ b/lib/video/postprocessing/visualmetrics/visualMetrics.js
@@ -111,8 +111,6 @@ module.exports = {
 
     scriptArgs.unshift(SCRIPT_PATH);
 
-    log.info("sparky")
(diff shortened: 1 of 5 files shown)


gmierz:
@soulgalore I realized that I have to make a couple more changes (removing the crop_region2 setting), but you can start testing and reviewing it now.

soulgalore:
@gmierz I did some quick tests on desktop and small viewports and it worked flawless for me and metrics looked good!

I'm gonna test with videos from Android + some more complicated sites than Wikipedia. I'm gonna add the new dependencies in the Docker containers so I can run the tests on the test server.

soulgalore:
I did testing on Android and it worked fine + built a Docker container and tested, fine too. For me that is enough as we turn it on with a flag. Then I can push a new version out for the tests that runs on https://dashboard.sitespeed.io and analyse the result.

gmierz:
That's great to hear! Thanks for the quick testing :)

@soulgalore, are you okay with the changes in the scripts? If so, I'll move it to visualmtrics-portable.py and add the flag.

soulgalore:
@gmierz yes please go ahead!

lib/video/postprocessing/visualmetrics/visualMetrics.js
```diff
@@ -16,9 +16,29 @@ const SCRIPT_PATH = path.join(
...
+  if (options.visualMetricsPortable) {
+    return PORTABLE_SCRIPT_PATH;
+  }
+  return SCRIPT_PATH;
+}
+
 module.exports = {
-  async checkDependencies() {
-    return execa(process.env.PYTHON || 'python', [SCRIPT_PATH, '--check']);
+  async checkDependencies(options) {
```
gmierz, line 37: @soulgalore is this method used anywhere? I wasn't able to find any references to it.

soulgalore:
@gmierz `npm run lint:fix` will fix that linting problem.

gmierz:
> @gmierz `npm run lint:fix` will fix that linting problem.

Thanks! I managed to fix it with that. However, I seem to be getting all sorts of other changes at the same time that are unrelated to my changes.

soulgalore:
Else if you don't get it work, do not spend time on it, do a happy merge and I can fix, I'm gonna add the dependencies in the GithHub actions at the same so we can have some tests running with the new functionality.

gmierz:
I've rebased it, but I think it might be something with my local setup since I also get it when I run it on the main branch on the newest commit. I'll look into it some other time as you suggest :)

gmierz:
@soulgalore all tests passed so I've merged it, thank you very much! :)

soulgalore:
Thank you @gmierz !!!!!

lib/video/postprocessing/visualmetrics/visualMetrics.js
```diff
@@ -16,9 +16,29 @@ const SCRIPT_PATH = path.join(
...
+  __dirname,
+  '..',
+  '..',
+  '..',
+  '..',
+  'browsertime',
+  'visualmetrics-portable.py'
+);
+
+async function getScript(options) {
```
soulgalore, line 29: FYI I changed this, that do not need to be async since we are not returning a promise or doing anything async.

PR #1745, opened 2022-03-30
Add Python 3.6+ requirement to visualmetrics-portable.py.

gmierz:
This patch adds a python 3.6+ requirement to the new `visualmetrics-portable.py` script (it doesn't work with Python 2, and has only been tested in 3.6+). It also fixes a small bug that can happen when `maxsize` is an Integer in the scale method (only used when `--contentful-video` is provided).


soulgalore:
@gmierz please feel free to merge fixes you do to that script yourself :)

gmierz:
Awesome, thanks @soulgalore!

PR #1749, opened 2022-03-31
Add a check for requirements before running.

gmierz:
This patch adds some code to check if the requirements have been installed. This prevents silent failures related to missing packages from slipping through.


soulgalore:
Hi @gmierz I think it needs to run the checks but do not output anything. We really on parsing the output from the visual metrics script and convert it to JSON and right now it outputs:

PR #1750, opened 2022-04-01
Only system out if checks fails

soulgalore:
See https://github.com/sitespeedio/browsertime/pull/1749#issuecomment-1085572842

### DIFF
--- a/browsertime/visualmetrics-portable.py
+++ b/browsertime/visualmetrics-portable.py
@@ -2286,54 +2286,52 @@ def calculate_hero_time(progress, directory, hero, viewport):
 def check_config():
     ok = True
 
-    print("ffmpeg:  ")
+    
     if get_decimate_filter() is not None:
-        print("OK")
+        logging.debug('FFMPEG found')
     else:
-        print("FAIL")
+        print("ffmpeg: FAIL")
         ok = False
 
-    print("Python 3.6+:  ")
+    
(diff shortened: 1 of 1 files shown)


gmierz:
@soulgalore I was going to ask to do this a bit differently but I really like the new behaviour that I'm seeing locally and it still works with our methodology. :)

When it's passing:
```
sparky@sparky-ThinkPad-X1-Carbon-6th:~/mozilla-source/mozilla-central$ ~/Downloads/visualmetrics-portable-new.py --check
sparky@sparky-ThinkPad-X1-Carbon-6th:~/mozilla-source/mozilla-central$
```

When it's failing:
```
sparky@sparky-ThinkPad-X1-Carbon-6th:~/mozilla-source/mozilla-central$ ~/Downloads/visualmetrics-portable-new.py --check
OpenCV-Python: FAIL
sparky@sparky-ThinkPad-X1-Carbon-6th:~/mozilla-source/mozilla-central$
```

soulgalore:
Cool, yeah we can do it better later, I just wanted to make it work for now.

gmierz:
> Cool, yeah we can do it better later, I just wanted to make it work for now.

It works perfectly now imo :)

That said, an ideal solution to me would be to remove this check_config method, remove all of the try/except statements that are scattered about, and move all the imports and the FFMPEG check to the top-level of the file outside of a function.

PR #1751, opened 2022-04-01
Remove unused functionality from the visual metrics portable script

soulgalore:
Remove the following:
* --gray (remove gray frames that was an old fix for a Chrome bug)
* --timeline that synced with trace from Chrome
* The functionality to use a video with multiple navigations and split that
* Remove white frames at the end
* Remove that the navigation could start white
* Remove the possibility to trim the end of the video
*  --findstart
* --viewporttime
* --white option
* --forceblank
* Gzip option for Python 2
* --histogram option
* Removing remove frames before first orange
* --maxframes
* --notification
* functionality to generate a new video from the frames

### DIFF
--- a/browsertime/visualmetrics-portable.py
+++ b/browsertime/visualmetrics-portable.py
@@ -405,9 +405,7 @@ def video_to_frames(
                     directories = [directory]
                     for dir in directories:
                         if orange_file is not None:
-                            remove_frames_before_orange(dir, orange_file)
                             remove_orange_frames(dir, orange_file)
-                        blank_first_frame(dir)
                         find_render_start(
                             dir, orange_file, cropped, is_mobile
                         )
@@ -512,32 +510,6 @@ def find_recording_platform(video):
             is_mobile = True
 
     return is_mobile
-
-def remove_frames_before_orange(directory, orange_file):
-    """Remove stray frames from the start of the video"""
-    frames = sorted(glob.glob(os.path.join(directory, "video-*.png")))
-    if len(frames):
-        # go through the first 20 frames and remove any that come before the first orange frame.
-        # iOS video capture starts with a blank white frame and then flips to
-        # orange before starting.
-        logging.debug("Scanning for non-orange frames...")
-        found_orange = False
-        remove_frames = []
-        frame_count = 0
-        for frame in frames:
-            frame_count += 1
-            if is_color_frame(frame, orange_file):
-                found_orange = True
-                break
-            if frame_count > 20:
-                break
-            remove_frames.append(frame)
-
-        if found_orange and len(remove_frames):
-            for frame in remove_frames:
-                logging.debug("Removing pre-orange frame %s", frame)
-                os.remove(frame)
-
 
 def remove_orange_frames(directory, orange_file):
     """Remove orange frames from the beginning of the video"""
--- a/browsertime/visualmetrics-portable.py
+++ b/browsertime/visualmetrics-portable.py
@@ -398,10 +398,6 @@ def video_to_frames(
                 gc.collect()
                 if extract_frames(video, directory, full_resolution, viewport):
                     client_viewport = None
-                    if find_viewport and options.notification:
-                        client_viewport = find_image_viewport(
-                            os.path.join(directory, "video-000000.png"), is_mobile
-                        )
                     directories = [directory]
                     for dir in directories:
                         if orange_file is not None:
@@ -411,7 +407,6 @@ def video_to_frames(
                         )
                         adjust_frame_times(dir)
                         eliminate_duplicate_frames(dir, cropped, is_mobile)
-                        eliminate_similar_frames(dir)
                         crop_viewport(dir)
                         gc.collect()
                 else:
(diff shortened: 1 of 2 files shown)


soulgalore:
@gmierz I'm gonna push more changes this weekend where I try to slimline the script. 

I have question: Do you use the script outside of Browsertime at Mozilla? I'm thinking about the ffprobe code and "knowing" if we test on mobile, when we run inside of browsertime that could be a flag (because we always know), that way we could probably sidestep using ffprobe?

gmierz:
With the new script, that sounds good! It was more difficult for us to do that with the old script because of how we were using it.

gmierz:
@soulgalore another thing you can get rid of is the Python2 checks. This script will only work with Python3.

soulgalore:
@gmierz I think I'm ready for a first look. Kind of shitty review with a lot of removes. I've been verifying that the functionality works and I removed things I know is not used by Browsertime. If this is ok for you, my next move is to tune some option so that they are used by default and then see if its meaningful to split the file into multiple files to make it easier to understand and add some better tests so we can run it standalone. It would be great to have a collection of videos that we test, where we test some edge cases.

gmierz:
Awesome! I'll review it on Monday.

Can we hold off on changing the default options and/or behaviours? I'd like to get this running in our CI before making more changes.

soulgalore:
Works great, I'll start by adding tests then.

gmierz:
@soulgalore I haven't forgotten about this! I'm just getting things working on our end with the existing script and fixing up some more bugs/issues while I'm at it.

gmierz:
I've triggered a set of tests here with this patch: https://treeherder.mozilla.org/jobs?repo=try&revision=29a21fd13047493de91c05bec84d9f374339b05a

gmierz:
@soulgalore this patch seems to have broken something for us. Here's a comparison of the metrics with and without your patch (the right side is data using this patch): https://treeherder.mozilla.org/perfherder/compare?originalProject=mozilla-central&originalRevision=2f240882d907fa71545869bf60a8a47aa69ae4d1&newProject=try&newRevision=29a21fd13047493de91c05bec84d9f374339b05a&framework=13&page=1&showOnlyComparable=1

browsertime/visualmetrics-portable.py
```diff
@@ -414,34 +403,15 @@ def video_to_frames(
...
-                        )
-                    if multiple and orange_file is not None:
-                        directories = split_videos(directory, orange_file)
-                    else:
-                        directories = [directory]
+                    directories = [directory]
                     for dir in directories:
-                        trim_video_end(dir, trim_end)
                         if orange_file is not None:
-                            remove_frames_before_orange(dir, orange_file)
```
gmierz, line 428: This method should stay or be combined with `remove_orange_frames`

gmierz (changes requested):

browsertime/visualmetrics-portable.py
soulgalore: Ok, I can add that but how do that work on your side? Usually the orange is added first, then the recording is started?

soulgalore: I've made the change but if I remember correctly the workflow is like this:
- Add the orange frame
- wait 
- start the video
 
I wonder than if that wait time is too short since it affected your tests, or is it another use case?

soulgalore: But I think that do not matter now, I added it so I hope it works when you retry :) Then we can look at the logic later.

gmierz:
New try run with the latest changes: https://treeherder.mozilla.org/jobs?repo=try&revision=dcfa4dc97e071d47ba8744e825c63308455c3563

soulgalore:
It looks like it worked right @gmierz ?

gmierz:
I had to retest because the last push had a lot of noisy data but the data looked really good in general. Only a couple metrics were 20-25% different but I think those are noise.

Here's a new test run for mobile: https://treeherder.mozilla.org/jobs?repo=try&revision=248e1237c260823f2ede62b9ec2b8b5f4114092c

Here's a new test run for desktop: https://treeherder.mozilla.org/#/jobs?repo=try&revision=79c8fa40c761824c7517f1f50abb4edb003c2811

I'll come back with a review by the end of my day.

gmierz:
Desktop looks great, here's a comparison of the numbers: https://treeherder.mozilla.org/perfherder/compare?originalProject=mozilla-central&originalRevision=f3aca6a6c4a43d3485a935380dee6c0e8a3b7fda&newProject=try&newRevision=79c8fa40c761824c7517f1f50abb4edb003c2811&framework=13&page=1&showOnlyComparable=1

The bigger differences are noise. Still waiting for android to finish.

browsertime/visualmetrics-portable.py
```diff
@@ -714,7 +634,7 @@ def find_video_viewport(
     video,
     directory,
     find_viewport,
-    viewport_time,
+
```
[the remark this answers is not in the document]
soulgalore: Thank, fixed!

gmierz:
Mobile tests have started completing and I don't see any major issues: https://treeherder.mozilla.org/perfherder/compare?originalProject=mozilla-central&originalRevision=f3aca6a6c4a43d3485a935380dee6c0e8a3b7fda&newProject=try&newRevision=248e1237c260823f2ede62b9ec2b8b5f4114092c&framework=13&page=12&showOnlyComparable=1

The larger differences are all noise.

gmierz (approved):

gmierz:
@soulgalore r+ thanks for doing this! :)

PR #1755, opened 2022-04-05
If Visual Metrics fail, store the original video.

soulgalore:
It's hacky but I think it will work and make it easier for us in the future. We can tune this later on.


soulgalore:
I'm pushing this so we can have it in the next release

PR #1759, opened 2022-04-06
Save the original video before getting rid of the orange frames.

gmierz:
This patch moves where the original video gets saved to before the orange frames are removed. There's a high chance that we won't be able to reproduce an issue if we remove these from the original video.

### DIFF
--- a/lib/video/video.js
+++ b/lib/video/video.js
@@ -104,16 +104,14 @@ class Video {
       );
     }
     if (this.options.video === true) {
-      if (this.options.tuneVideo === true) {
-        await finetuneVideo(
-          this.videoDir,
-          this.videoPath,
-          this.index,
-          videoMetrics,
-          timingMetrics,
-          this.options
-        );
-      }
+      await finetuneVideo(
+        this.videoDir,
+        this.videoPath,
+        this.index,
+        videoMetrics,
+        timingMetrics,
+        this.options
+      );
     } else {
       await fileUtil.removeFile(this.videoPath);
     }
--- a/lib/video/video.js
+++ b/lib/video/video.js
@@ -104,14 +104,16 @@ class Video {
       );
     }
     if (this.options.video === true) {
-      await finetuneVideo(
-        this.videoDir,
-        this.videoPath,
-        this.index,
-        videoMetrics,
-        timingMetrics,
-        this.options
-      );
+      if (this.options.tuneVideo === true) {
+        await finetuneVideo(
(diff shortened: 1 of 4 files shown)


gmierz:
@soulgalore I'm also adding an option to disable the fine tune video. It's breaking things for me locally because there seems to be a run condition happening often there where visual-metrics are being processed, and at the same time the video is being fine tuned.

soulgalore:
I looked at the code and I think we should try to fix that "removing orange start"- code, so that the promise is fulfilled when the new file is done, I don't think that happens now and that's why you get that problem. Let me have a go at that tomorrow.

soulgalore:
About that error in the Linux container, I think that started to show up with Firefox 99? I tried to reproduce it locally on my Mac but no luck so far.

lib/video/video.js
```diff
@@ -104,14 +104,16 @@ class Video {
...
     if (this.options.video === true) {
-      await finetuneVideo(
-        this.videoDir,
-        this.videoPath,
-        this.index,
-        videoMetrics,
-        timingMetrics,
-        this.options
-      );
+      if (this.options.tuneVideo === true) {
```
soulgalore, line 107: Like it is now this will have other complications? By default we add the time and then also some metrics ontop of the video. If the orange is there, the timer is not insync. We also add the metrics in the video, when it happens.

gmierz: I'm not sure I understand the issue you're pointing to. However, I can remove this flag now that the issue I was having was resolved and that the original videos are now true originals.

soulgalore: Sorry, I'll try to explain better: if we add a flag where you as a user can choose not to tune the video, the video will by default have overlay with a timer and the metrics. But that timer will not be correct (=matching the video), the same with the metric. If we add that functionality, I think we also need to make sure we do not add the timer and text for the metrics.

gmierz: Ah I see, so I'm confused about this because it looks like adding the timer is done within this `fineTuneVideo` method so I assumed that flag would skip adding the timer as well.

gmierz: I've since remove the flag though because the rename to copyFile change I made fixed the issue.

soulgalore: I see I could be wrong, I think the video code can be rewritten so its easier to follow.

soulgalore:
Thank you @gmierz !

gmierz:
Thanks for the review @soulgalore :)

PR #1762, opened 2022-04-08
Test on NodeJS 12, 14, 16

soulgalore:


soulgalore:
@gmierz I added so we run the unit test with NodeJS 12, 14, 16 and if possible I want to remove 12 when it reached EOL at the end of April. Let me know what you think :)

soulgalore:
So it fails in 12.x.

Issue #1753, opened 2022-04-04
visualmetrics-portable hero time error


soulgalore:
I picked up one error on the test server from the new portable script:

```
09:48:36.736 - '<' not supported between instances of 'NoneType' and 'int'
Traceback (most recent call last):
  File "/usr/src/app/node_modules/browsertime/browsertime/visualmetrics-portable.py", line 2707, in main
    metrics = calculate_visual_metrics(
  File "/usr/src/app/node_modules/browsertime/browsertime/visualmetrics-portable.py", line 1957, in calculate_visual_metrics
    hero_timings_sorted = sorted(
TypeError: '<' not supported between instances of 'NoneType' and 'int'

```

I'll try to get a video for that tonight.

gmierz:
Thanks for finding this! I'll take a look once I have the video and visual elements that caused this.

soulgalore:
I couldn't reproduce this locally and looking at the test result it wasn't uploaded, so probably something was fundamental broken when we tried to record the video. Gonna close it now and then create a new one if it happens again. One thing that would be cool that we could implement is that we keep the original video if visual metrics fails, so its easy to try locally.

gmierz:
+1 to keeping the original video that failed, that's a good idea

soulgalore:
It's the same page again but a different error:

```
15:26:12.059 - tile cannot extend outside image
Traceback (most recent call last):
  File "/usr/src/app/node_modules/browsertime/browsertime/visualmetrics-portable.py", line 111, in crop_im
    return Image.fromarray(
  File "/usr/local/lib/python3.8/dist-packages/PIL/Image.py", line 2834, in fromarray
    return frombuffer(mode, size, obj, "raw", rawmode, 0, 1)
  File "/usr/local/lib/python3.8/dist-packages/PIL/Image.py", line 2761, in frombuffer
    return frombytes(mode, size, data, decoder_name, args)
  File "/usr/local/lib/python3.8/dist-packages/PIL/Image.py", line 2707, in frombytes
    im.frombytes(data, decoder_name, args)
  File "/usr/local/lib/python3.8/dist-packages/PIL/Image.py", line 780, in frombytes
    d.setimage(self.im)
ValueError: tile cannot extend outside image
```

I'll push so we store the original video and keep this open until we have video we can work with.

gmierz:
Sounds good, I think that either the image or the crop region being given is empty. Can you get the visual element positions too? I had a visualElements file last time I tested it

[...]

PR #1770, opened 2022-04-19
Issue #1753: Handle negative x/y offsets when cropping images.

gmierz:
This patch fixes an issue where the hero element has a negative x/y offset that was not being handled correctly. With this change, when a negative offset causes the starting indices to go negative we set the base position to 0.


gmierz:
See issue here: https://github.com/sitespeedio/browsertime/issues/1753

gmierz:
I'll merge this as it only modifies the portable vismet processing script and doesn't affect anything else.

soulgalore:
Thank you @gmierz !

PR #1780, opened 2022-05-04
Add an acceptable error of 5 pixels to last frame matches.

gmierz:
This patch adds an extra 5 pixel buffer to the last frame matches to handle an issue where the frames previously matched but are currently mismatched by a few pixels which causes regressions in the LastVisualChange metric.

### DIFF
--- a/browsertime/visualmetrics-portable.py
+++ b/browsertime/visualmetrics-portable.py
@@ -1129,7 +1129,7 @@ def eliminate_duplicate_frames(directory, cropped, is_mobile):
                 baseline = files[0]
                 previous_frame = baseline
                 for i in range(1, count):
-                    if frames_match(baseline, files[i], 15, 0, crop, None):
+                    if frames_match(baseline, files[i], 15, 5, crop, None):
                         if previous_frame is baseline:
                             duplicates.append(previous_frame)
                         else:


soulgalore:
Ah great @gmierz ! Actually I was working on an issue, I could see that when we pushed Firefox 100, last visual change increased, but I couldn't see anything in the video nor the filmstrip. I've turned on the full filmstrip view to try to understand better what's going on, but hopefully that is not needed then :) One thing though, I'm thinking in the future we should add a way of setting the filmstrip screenshot size, the default one is actually too small when you try to see those little changes. And the full screenshot is really large.

gmierz:
Good idea about the filmstrip size! @soulgalore could you add that to the rewrite vismet issue or create a new issue for it?

And thanks for the merge :)
Did it resolve the issue you were having?

soulgalore:
I'll push it tonight and will know tomorrow morning :)

gmierz:
Sounds good, it fixed the issues we were having without introducing any new regressions so far: https://treeherder.mozilla.org/perfherder/compare?originalProject=mozilla-central&originalRevision=9854360e66cc3f82d3e34af96189fdd92b54653e&newProject=try&newRevision=c5a739d6883f407455a149b593031fd3372ab06a&framework=13&page=1&showOnlyComparable=1&showOnlyImportant=1

The changes that I'm seeing there are noise based on the historical graphs.

soulgalore:
@gmierz  here's a screenshot for one of the URLs where I look at LastVisualChange:
ff100

That 5 pixel change took it down to the new normal with Firefox 100. It look (at least on that test server) that Firefox 100 introduced higher last visual change and fully loaded (when all request/responses are done).

I think though that 5 pixel maybe is too small? Check these two screenshots:
ms_002767
ms_002700

You can see the full film strip here:
https://s3.amazonaws.com/results.sitespeed.io/politicalparties/2022-05-05-04-50-26/pages/moderaterna_se/index.html#filmstrip

gmierz:
@soulgalore thanks, I'm glad to see you have a graph available for this!

Let's wait to see if there are any more issues before increasing it further.

If it's still an issue, I think we should instead add some logic for determining if the different pixels are scattered or not. This would let us determine if the (for example) 10 different pixels are arranged in a square representing a thumbnail on the page, or if they're just noise scattered around the image. It should be relatively easy to implement this with an image segmentation algorithm (e.g. nearest-neighbours).

soulgalore:
@gmierz cool, let me create issue for nearest-neighbours. I have more URLs where this is a problem + seen this before with the old script also, for example has fonts on Linux been a problem for Chrome where anti alias sometimes is added as a last step (it seems).

soulgalore:
But what was strange is that Firefox 100 seems to trigger this or at least have later visual change.

gmierz:
Based on the graph you provided, it looks like FF100 introduced a regression to the LVC but the massive spike was something else that was fixed by this patch. Or am I misunderstanding something? Do you also see this issue with the standard vismet processing script?

soulgalore:
Yes that is correct. The patch fixed some big problem, however the screenshots I posted is after that patch so there are still some fixes that can be done.

At work we use the old script and I could see the same thing there when we pushed FF 100, this is the Wikipedia Barack Obama page:
ff100

I'm gonna create an issue for that but haven't had time yet, I want to collect some more data first.

gmierz:
Ah ok, thanks for clarifying! I agree, filling a bug on bugzilla for this would be a good idea (please cc me on it).

Issue #1784, opened 2022-05-09
visualmetrics-portable throws error for hero times


soulgalore:
I'll try to hunt down the original video and settings for it.

```
 tile cannot extend outside image
Traceback (most recent call last):
  File "/usr/src/app/node_modules/browsertime/browsertime/visualmetrics-portable.py", line 116, in crop_im
    return Image.fromarray(
  File "/usr/local/lib/python3.8/dist-packages/PIL/Image.py", line 2949, in fromarray
    return frombuffer(mode, size, obj, "raw", rawmode, 0, 1)
  File "/usr/local/lib/python3.8/dist-packages/PIL/Image.py", line 2876, in frombuffer
    return frombytes(mode, size, data, decoder_name, args)
  File "/usr/local/lib/python3.8/dist-packages/PIL/Image.py", line 2822, in frombytes
    im.frombytes(data, decoder_name, args)
  File "/usr/local/lib/python3.8/dist-packages/PIL/Image.py", line 827, in frombytes
    d.setimage(self.im)
ValueError: tile cannot extend outside image
05:54:16.768 - 'NoneType' object has no attribute 'save'
Traceback (most recent call last):
  File "/usr/src/app/node_modules/browsertime/browsertime/visualmetrics-portable.py", line 2234, in calculate_hero_time
    target_mask.save(target_mask_path)
AttributeError: 'NoneType' object has no attribute 'save'
05:54:16.768 - '<' not supported between instances of 'NoneType' and 'int'
Traceback (most recent call last):
  File "/usr/src/app/node_modules/browsertime/browsertime/visualmetrics-portable.py", line 2725, in main
    metrics = calculate_visual_metrics(
  File "/usr/src/app/node_modules/browsertime/browsertime/visualmetrics-portable.py", line 1962, in calculate_visual_metrics
    hero_timings_sorted = sorted(
TypeError: '<' not supported between instances of 'NoneType' and 'int'

```

soulgalore:
It's reproducible using:
` ./visualmetrics-portable.py --herodata 1-visualElements.json.gz.failed.json.gz -i 1.mp4.failed.mp4 --orange --force --renderignore 5 --json --viewport --viewportretries 60 --viewportminheight 100 --viewportminwidth 100  -vvvv`

[1-visualElements.json.gz.failed.json.gz](https://github.com/sitespeedio/browsertime/files/8658746/1-visualElements.json.gz.failed.json.gz)

[...]

PR #1788, opened 2022-05-12
Handle all boundaries when cropping image in portable script.

gmierz:
This patch changes the `crop_im` method to handle all boundaries (going past the edges in either direction). It also throws an exception when an empty image is produced as this means that the crop region being given is incorrect.

See Issue #1784.


soulgalore:
Great thank you @gmierz

PR #1790, opened 2022-05-17
Add a debug mode to browsertime with abitrary pausing.

gmierz:
This patch adds a --debug flag that can be used to enter debug mode. This mode enables a hotkey `ctrl+alt+g` which can be used to pause execution after the iteration is complete. It's also exposed in the test script so it can be used by custom test scripts as they may perform their own iterations.

Furthermore, when in --debug mode the Firefox browser will be detached from the terminal and open the devtools panel by default.

### DIFF
--- a/lib/firefox/webdriver/builder.js
+++ b/lib/firefox/webdriver/builder.js
@@ -199,6 +199,11 @@ module.exports.configureBuilder = function (builder, baseDir, options) {
     }
   }
 
+  if (options.debug) {
+    ffOptions.addArguments("--devtools");
+    ffOptions.setPreference("detach", true);
+  }
+
   if (options.headless) {
     ffOptions.headless();
   }
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "lodash.pick": "4.4.0",
     "lodash.set": "4.3.2",
     "selenium-webdriver": "4.1.2",
+    "uiohook-napi": "^1.1.0",
     "yargs": "17.4.1"
   },
   "optionalDependencies": {
--- a/package.json
+++ b/package.json
@@ -19,7 +19,6 @@
     "find-up": "5.0.0",
     "get-port": "5.1.1",
     "hasbin": "1.2.3",
-    "hotkeys-js": "^3.9.3",
     "intel": "1.2.0",
     "lodash.get": "4.4.2",
     "lodash.groupby": "4.6.0",
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "find-up": "5.0.0",
     "get-port": "5.1.1",
     "hasbin": "1.2.3",
+    "hotkeys-js": "^3.9.3",
     "intel": "1.2.0",
     "lodash.get": "4.4.2",
     "lodash.groupby": "4.6.0",
@@ -38,8 +39,8 @@
     "eslint-config-prettier": "8.5.0",
     "eslint-plugin-prettier": "4.0.0",
     "prettier": "2.6.2",
-    "serve-handler": "6.1.3",
-    "serve": "13.0.2"
+    "serve": "13.0.2",
+    "serve-handler": "6.1.3"
   },
   "engines": {
     "node": ">=14.19.1"
--- a/lib/core/engine/iteration.js
+++ b/lib/core/engine/iteration.js
@@ -1,5 +1,7 @@
 'use strict';
 
+const { hotkeys } = require('hotkeys-js');
+
 const webdriver = require('selenium-webdriver');
 const log = require('intel').getLogger('browsertime');
 const SeleniumRunner = require('../seleniumRunner');
@@ -244,7 +246,11 @@ class Iteration {
       }
 
       await engineDelegate.beforeBrowserStop(browser, index, result);
-      await browser.stop();
+
+      if (!options.debug) {
+        await browser.stop();
+      }
+
       // Give the browsers some time to stop
       await delay(2000);
       await engineDelegate.afterBrowserStopped();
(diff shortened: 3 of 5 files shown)


soulgalore:
Thank you @gmierz this looks super cool! I'll will test it out later this week!

package.json
```diff
@@ -27,6 +27,7 @@
     "lodash.pick": "4.4.0",
     "lodash.set": "4.3.2",
     "selenium-webdriver": "4.1.2",
+    "uiohook-napi": "^1.1.0",
```
soulgalore, line 30: I didn't get this to work on my Mac and it also seems to fail on the GitHub Action on Mac? I'm thinking two things:
* Maybe we could have it as an optional dependency (and follow the pattern we have for that)? But it should still work on all platforms, do you know if it should? I could also see in the package.json for that package that it seems to be NodeJS 16 and never.
* Lets sync how we should communicate between Browsertime and the browser. If we can't get keys to work, maybe we could use devtools and global variables in JS?

lib/firefox/webdriver/builder.js
```diff
@@ -199,6 +199,11 @@ module.exports.configureBuilder = function (builder, baseDir, options) {
     }
   }
 
+  if (options.debug) {
```
soulgalore, line 202: Think we need to check for not running on Android? Or do the devtools switch work on Android?

soulgalore: Also I can add the same for Chrome.

soulgalore (changes requested):
Check if you can get uiOhook to work, then I can try it some more!

package.json
gmierz: Ah good catch! I was running in node-16 so maybe it pulled a version that only works for node-16.

gmierz:
Thanks for the review @soulgalore :) I'll update the PR later next week

soulgalore:
@gmierz I'm really found of this PR and I can help out getting things to work, but I need your help/guidance how we gonna do with the uiohook-nap library since it it fails to run on Mac (maybe others too)? One way could be do do this in multiple steps and first skip that lib and just communicate using devtools (or at least try it).

gmierz:
@soulgalore regarding the new package, this was the only one I could find that worked for me. There are others but they are either broken, or only run on the client side of node (we can't use it within browsertime code as they require the DOM).

Can you elaborate on your idea with using devtools?

soulgalore:
Ok! Let me try to make a POC, hope to have it ready Friday/Saturday.

soulgalore:
@gmierz ok I tried now and maybe it's a different use case of what you try to do?

I've changed so that Firefox isn't detached from browsertime and then I added a debug command so that in your script where you want to debug it you add:
` await commands.debug(); ` (or pause or whatever) and then it sets browsertime.debug = true and loops until it's false. When my debug is finished, I set `browsertime.debug = false` and the script continous to the next pause/breakpoint in the script. 

To fix when we test one URL I guess we could add that command after the first URL is finished.

However it's not as fancy as stop/resume by clicking a button?

soulgalore:
Another alternative could be for browser that supports it to load an extension when we run in debug mode that adds a continue button, so that you don't have to add JS directly in the console, I guess that would work?

gmierz:
@soulgalore I think the devtools method you mentioned would work well, but we would need to keep Firefox detached from the terminal so that the browser doesn't close if the terminal process is killed. I think we can do that with your method as well though so there's no issues.

The method is very similar to what I had implemented and the only real difference is the lack of hotkeys. Using hotkeys, we have the same behaviour but it feels "fancier". We could also leave a debug loop after the iteration ends but before the browser stops and allow a user to continue by pressing enter in the console (this can be done easily with readline - it's global hotkeys that are an issue). This would result in almost exactly the same behaviour as we have with hotkeys but it would be slightly less dynamic which isn't a problem to me.

soulgalore:
> but we would need to keep Firefox detached from the terminal so that the browser doesn't close if the terminal process is killed.

I didn't fully understand that part, how is that used? Like do you as a dev close the command line? I'm thinking now with the timeout fix you did the command line will just wait? 

Ok cool, if we can get rid of uiohook-napi so it can run on all platforms that's the most important for me. Do you want me to continue and can I push to your branch? You readline idea sounds good too!

gmierz:
> > but we would need to keep Firefox detached from the terminal so that the browser doesn't close if the terminal process is killed.
> 
> I didn't fully understand that part, how is that used? Like do you as a dev close the command line? I'm thinking now with the timeout fix you did the command line will just wait?

Not the command line/terminal, but the browsertime node process. If someone were to exit that browsertime process, firefox would stay open when running in detached mode. Right now, if we kill browsertime, it also kills the browser session.

> 
> Ok cool, if we can get rid of uiohook-napi so it can run on all platforms that's the most important for me. Do you want me to continue and can I push to your branch? You readline idea sounds good too!

Yes please if you have some time! Otherwise, I'll get back to it later next week.

soulgalore:
> Yes please if you have some time! Otherwise, I'll get back to it later next week

Ok, I'll try to do something this weekend then.

> Not the command line/terminal, but the browsertime node process. If someone were to exit that browsertime process, firefox would stay open when running in detached mode. Right now, if we kill browsertime, it also kills the browser session.

Yeah I got that but I don't understand why someone kill that process when you debug? :)

gmierz:
Oh sorry! I'll find more information for you next week about this, but it's more for browser rather than web development. We do this in all of our existing test harnesses at Mozilla.

soulgalore:
Pushed https://github.com/sitespeedio/browsertime/pull/1798 as the alternative

Issue #1765, opened 2022-04-13
EBUSY error on Windows using '--firefox.collectMozLog'


bthrall:
### Have you read the documentation?

- [X] Yes, I've read the [how to make a reproducable bug guide](https://www.sitespeed.io/documentation/sitespeed.io/bug-report/)
- [X] Yes, I've read the [how to debug my script guide](https://www.sitespeed.io/documentation/sitespeed.io/scripting/#debug)

### URL

https://www.reddit.com

### What are you trying to accomplish

I am trying to capture Firefox logs so I can debug my performance test, but when I do, Browsertime fails on Windows with 'EBUSY' while trying to rename the Firefox log files.

Because Browsertime is failing to move the log file, the entire run fails and I get no results in _browsertime.json_.

Removing the `--firefox.collectMozLog` option avoids the problem, but I also don't get logs from Firefox.

Browsertime version is 16.0.1

### What browser did you use?

Firefox

### How to reproduce

```shell
browsertime -b firefox --firefox.collectMozLog --firefox.setMozLog "ScriptLoader:5" -n 1 https://www.reddit.com
```

### Relevant log output

```shell
[2022-04-13 10:14:25] INFO: Running tests using Firefox - 1 iteration(s)
[2022-04-13 10:14:28] INFO: Testing url https://www.reddit.com iteration 1
[2022-04-13 10:14:41] ERROR: Error: EBUSY: resource busy or locked, rename 'C:\Users\Bryan Thrall\Documents\performance\browsertime-results\www.reddit.com\2022-04-13T101425-0500\moz_log.txt.child-1.moz_log' -> 'C:\Users\Bryan Thrall\Documents\performance\browsertime-results\www.reddit.com\2022-04-13T101425-0500\pages\www_reddit_com\data\moz_log.txt.child-1.moz_log-1.txt''
[2022-04-13 10:14:41] INFO: https://www.reddit.com
[2022-04-13 10:14:41] INFO: Wrote data to browsertime-results\www.reddit.com\2022-04-13T101425-0500
```

soulgalore:
Hi @bthrall thanks for creating the issue! I pushed a test if the Windows GitHub action to check if I can reproduce it.

soulgalore:
Yep I got the same. I think we rename those file when Firefox is still running, I'll make a PR to fix that early next week.

soulgalore:
I think this has been fixed.

PR #1802, opened 2022-06-07
Move mozlog collection to postWork method.

gmierz:
This patch moves the Firefox mozlog collection to the postWork method. This fixes an issue where we were trying to move the moz logs while they were still being written to. This issue exists on Linux, and Mac as well but we don't get an error there when we do the renaming/moving.

This fixes issue #1765.


soulgalore:
Thank you @gmierz

PR #1803, opened 2022-06-07
Update wait.js with byCondition method

Icecold777:
adds wait.byCondition where sitespeed.io waits for any expression which eventually returns true or false(error will be thrown)


soulgalore:
Hi @Icecold777 ah cool, thanks for sharing, I will have a look the coming days!

soulgalore:
Thank you @Icecold777 ! Think this will come in the sitespeed.io release that will go out next weekend.

PR #1805, opened 2022-06-10
Add extra delay when Firefox Window recorder stops

soulgalore:
If the Firefox Window recorder is used in https://bugzilla.mozilla.org/show_bug.cgi?id=1773127 this could potentially fix that. If that's the case, there's something in the code in the window recorder that do not work as it should, since the video should stop before we start to collect metrics.

### DIFF
--- a/lib/video/screenRecording/firefox/firefoxWindowRecorder.js
+++ b/lib/video/screenRecording/firefox/firefoxWindowRecorder.js
@@ -7,6 +7,7 @@ const path = require('path');
 const execa = require('execa');
 const { removeDirAndFiles } = require('../../../support/fileUtil');
 const { Android } = require('../../../android');
+const delay = ms => new Promise(res => setTimeout(res, ms));
 
 const unlink = util.promisify(fs.unlink);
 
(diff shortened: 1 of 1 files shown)


lib/video/screenRecording/firefox/firefoxWindowRecorder.js
```diff
@@ -174,6 +175,9 @@ module.exports = class FirefoxWindowRecorder {
     log.info('Stop firefox window recorder.');
     await enableWindowRecorder(false, this.browser);
 
+    // Add some extra delay to make sure the recording is finished.
+    await delay(this.options.firefox.windowRecorderStopDelay || 1000);
```
gmierz, line 179: There is also a delay for Android video recordings: https://github.com/sitespeedio/browsertime/blob/9004781fda367edcbd7285723949a02c23ec998f/lib/video/screenRecording/android/recorder.js#L14

Would it be possible to generalize these delays?

soulgalore: Sounds good, but lets wait first if there's way to fix the window recorder. We could probably rework both timoutes and wait times to just make it simpler.

soulgalore:
Lets close this since I haven't heard that is needed :)

PR #1809, opened 2022-06-14
Exit after maxLoadTime if the loadEventEnd do not happen.

soulgalore:
We got users (and I've seen this myself on the test server) that for some URLs we hit the max timeout time configured for the pageCompleteCheck (5 minutes), that means we get an error and the test fails. However it seems that some sites never fires the loadEventEnd, so this fix ends the tests after 2 minutes (or configurable) if loadEventEnd do not happen.

@gmierz I don't think this will have any effect on your tests, but wanted to verify that I'm on the right path. I do wanna cleanup with with all those timouts but I'll save that for after the summer.

### DIFF
--- a/lib/core/seleniumRunner.js
+++ b/lib/core/seleniumRunner.js
@@ -143,6 +143,7 @@ class SeleniumRunner {
    */
   async wait(pageCompleteCheck, url) {
     const waitTime = this.options.pageCompleteWaitTime || 5000;
+    const maxLoadTime = this.options.pageMaxLoadTime || 2 * 60 * 1000;
     if (!pageCompleteCheck) {
       pageCompleteCheck = this.options.pageCompleteCheckInactivity
         ? pageCompleteCheckByInactivity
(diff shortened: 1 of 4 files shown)


lib/core/seleniumRunner.js
```diff
@@ -160,7 +161,7 @@ class SeleniumRunner {
         'for page complete check script to return true',
         function (d) {
           return d
-            .executeScript(pageCompleteCheck, waitTime)
+            .executeScript(pageCompleteCheck, { waitTime, maxLoadTime })
```
gmierz, line 164: It would be cool to see the `maxLoadTime` used in all checks since 2 minutes seems really long for any pageload no matter how you check the end of it. It might be easy to do that by having a separate `maxLoadTime` check just before this check.

gmierz (approved):
Looks good! Just a small idea for you summer rework in the comments.

lib/core/seleniumRunner.js
soulgalore: Yep, let me fix that, good idea.

soulgalore:
Closing since https://github.com/sitespeedio/browsertime/pull/1810 will do the job better (and do not need to be a major release).

PR #1814, opened 2022-06-27
It seems FF again has a HAR issue

soulgalore:


soulgalore:
No that did not help.

PR #1821, opened 2022-08-03
Revert the orange frame workaround we did for Firefox

sefeng211:
In https://bugzilla.mozilla.org/show_bug.cgi?id=1606365,
we did a workaround to fix a Firefox issue, and
this issue is no longer persist, so we can remove this changes.

### DIFF
--- a/lib/video/screenRecording/setOrangeBackground.js
+++ b/lib/video/screenRecording/setOrangeBackground.js
@@ -2,7 +2,7 @@
 
 const log = require('intel').getLogger('browsertime.video');
 const { until, By } = require('selenium-webdriver');
-module.exports = async function (driver, options) {
+module.exports = async function (driver) {
   log.debug('Add orange color');
   // We tried other ways for Android (access an orange page)
   // That works fine ... but break scripts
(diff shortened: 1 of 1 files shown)


sefeng211:
Hmm, there's a test failure on Firefox, @soulgalore is this legit? 
```
Error: 8-03 15:30:14][ERROR] Got an error from HAR Export Trigger {"error":"The HAR export trigger returned an empty HAR{\"version\":\"1.2\",\"creator\":{\"name\":\"Firefox\",\"version\":\"103.0.1\"},\"browser\":{\"name\":\"Firefox\",\"version\":\"103.0.1\"},\"pages\":[],\"entries\":[]}"}
  ✖ engineTest › Load one URL with none page load strategy Rejected promise returned by test
Error: 8-03 15:30:14][ERROR] TypeError: Cannot read properties of undefined (reading 'log')
```

soulgalore:
Yep it is legit in the way that something is wrong but it do not depend on your change. That error started to happen a couple of Firefox releases again but mostly on GitHub actions and not everytime.

soulgalore:
@sefeng211 I'll be back home on Friday and can merge then but feel free to self merge before that if you want. Thank you for creating the revert :)

soulgalore:
thank you @sefeng211 !

PR #1823, opened 2022-08-06
Bump to selenium 4.3.1 and chrome-remote-interface-0.31.3

soulgalore:


soulgalore:
Ah Selenium broke this upgrade with a change in 4.3.1 in https://github.com/SeleniumHQ/selenium/pull/10796 - lets see when I have time have a look at it :(

PR #1848, opened 2022-09-20
Add ability to press the home button on Android before a test starts

soulgalore:
Use `--androidPretestPressHomeButton`.


soulgalore:
Ooops, this was already pushed.

PR #1850, opened 2022-09-23
Add ability to ignore shutdown failures on android.

gmierz:
This patch adds an option to ignore shutdown failures when running on android.


soulgalore:
Thank you @gmierz let me fix those lint errors.

PR #1854, opened 2022-10-08
Disable password popup on Android.

soulgalore:
There was an old comment that this only work on Desktop but is that really the case? Lets try it out because the password popup is annoying.

Let me test this when I get a hold of a phone :)


soulgalore:
Running on Android trying to set those prefs I get:
` invalid argument: cannot parse capability: goog:chromeOptions from invalid argument: unrecognized chrome option: prefs`

soulgalore:
See https://monorail-prod.appspot.com/p/chromedriver/issues/detail?id=1206

PR #1874, opened 2022-11-16
Provide active tab id to gecko profiler StartProfiler API

canova:
In the `Services.profiler.StartProfiler` API we have an optional parameter to pass the active tab id. It is the identifier for browser's active tab so the profiler can determine which website was inside the browser's active tab during the profiling session.

This is optional because it's not a crucial profiler feature, but it's useful to provide this.
You can see the code similar to this inside Firefox: https://searchfox.org/mozilla-central/rev/d7d2cc647772de15c4c5aa47f74d25d0e379e404/devtools/client/performance-new/popup/background.jsm.js#382-393 and
https://searchfox.org/mozilla-central/rev/d7d2cc647772de15c4c5aa47f74d25d0e379e404/devtools/shared/performance-new/recording-utils.js#16-34

While I was changing this part, I also added some more context in the commments for the API changes that we've done in the past. Hope it's useful!

To test this change locally, you can run it with:
```
node bin/browsertime.js https://sitespeed.io -b firefox --firefox.geckoProfiler --iterations 1
```
and it should produce the profile data inside `<browsetime-results-dir>/pages/sitespeed_io/data/geckoProfile-1.json.gz`.
If you search for `activeTabID` field inside that json, you should see a non-zero value. Previously this field was always zero.

### DIFF
--- a/lib/firefox/geckoProfiler.js
+++ b/lib/firefox/geckoProfiler.js
@@ -73,14 +73,38 @@ class GeckoProfiler {
     );
 
     let script;
-    if (parameterLength === 7) {
-      script = `Services.profiler.StartProfiler(${bufferSize},${interval},${featureString},
-        ${chosenFeatures.length},${threadString},${chosenThreads.length});`;
-    } else if (parameterLength === 5 || parameterLength === 6) {
-      script = `Services.profiler.StartProfiler(${bufferSize},${interval},${featureString},${threadString});`;
-    } else {
-      log.error('Unknown Gecko Profiler API');
-      throw new Error('Unknown Gecko Profiler API');
+    switch (parameterLength) {
+      case 7: {
+        // Oldest profiler API. It was deprecated in Firefox 69.
+        script = `Services.profiler.StartProfiler(${bufferSize},${interval},${featureString},
+          ${chosenFeatures.length},${threadString},${chosenThreads.length});`;
+        break;
+      }
+      case 6: {
+        // Newest profiler API. It's being used since Firefox 72.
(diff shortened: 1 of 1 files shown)


canova:
Interesting. CI shows that a unit test has failed but I can't reproduce it locally.

soulgalore:
Hi @canova ok cool. I did retry. The thing is that when you run with the profiler turned on using cloud instances (the CI or for example on AWS or Digital Cloud) there's quite often failure like this. I wonder if there could be a limit on memory/CPU or something? However been running it and seen the same on bare metal servers where it just fails on timeouts.

canova:
> Hi @canova ok cool. I did retry. The thing is that when you run with the profiler turned on using cloud instances (the CI or for example on AWS or Digital Cloud) there's quite often failure like this. I wonder if there could be a limit on memory/CPU or something? However been running it and seen the same on bare metal servers where it just fails on timeouts.

Thanks for the retry!

We have some intermittent failures on our CI time to time too I think. It's something we can investigate more. We can maybe try to reduce the profiler buffer size more (current default is ~100MB) for the tests, so it doesn't restrain the CI memory.

But are we running the gecko profiler inside the unit tests at all? I couldn't find any gecko profiler related unit test in the codebase.  There is only one geckoProfiler run in [linux.yml file](https://github.com/sitespeedio/browsertime/blob/31402fc7ba7ada5ff470e9c0302a0a14f1d74a53/.github/workflows/linux.yml#L64-L65). I think that might be unrelated.

canova:
Btw, is there anything I need to do for this PR? I think it should be ready for review.

soulgalore:
Hi @canova sorry I forgot, let me fix it today.

PR #1875, opened 2022-11-18
Get CPU metrics using Glean

afinder-mozilla:
Using Glean to extract CPU/Power metrics

### DIFF
--- a/lib/firefox/webdriver/firefox.js
+++ b/lib/firefox/webdriver/firefox.js
@@ -73,6 +73,14 @@ class Firefox {
    * Before each URL/test runs.
    */
   async beforeEachURL(runner) {
+    await runner.runPrivilegedScript(`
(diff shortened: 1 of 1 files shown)


lib/firefox/webdriver/firefox.js
```diff
@@ -73,6 +73,14 @@ class Firefox {
    * Before each URL/test runs.
    */
   async beforeEachURL(runner) {
+    await runner.runPrivilegedScript(`
```
soulgalore, line 76: Here you can just return the promise instead of await. I'll fix that when I merged :)

soulgalore:
`npm run lint:fix` fixes the linting (just so you know for the next time!).

afinder-mozilla:
@soulgalore Thank you for merging the PR Peter!

PR #1876, opened 2022-11-25
Use Python3 as default

soulgalore:
Since visualmetrics-portable have` #!/usr/bin/env python3` I'm thinking maybe we should use python3 as default or how do others treat python/python3, @gmierz do you know what's best practice?

### DIFF
--- a/lib/video/postprocessing/visualmetrics/visualMetrics.js
+++ b/lib/video/postprocessing/visualmetrics/visualMetrics.js
@@ -37,7 +37,10 @@ function getScript(options) {
 }
 
 export async function checkDependencies(options) {
-  return execa(process.env.PYTHON || 'python', [getScript(options), '--check']);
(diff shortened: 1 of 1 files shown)


gmierz:
I agree with making it default but I think what you had before was better. That'll select the user's default python installation which should be python 3 because python 2 is deprecated now. Maybe you could add a python3 check early on to force it?

Alternatively, you could check python, then python3 as a backup. On windows we also have `py -3`.

soulgalore:
The thing is it doesn't work by default on a new Mac I think? I don't think Python is defaulting to Python3, but lets see if I ever need to reinstall Mac OS :D

gmierz:
I think it does default to python3 now. The new engineers on our team got macs and they were at python 3.8 when they started. I'd be very surprised if python2 was still a default on macs.

PR #1896, opened 2023-02-05
Add a timeout to gecko profiler gzipping phase

canova:
In our performance testing CI in Firefox, it looks like browsertime is hanging right after the `"Stop GeckoProfiler."` log intermittently. We are not sure where exactly it's hanging but I strongly suspect that it's hanging at the `storageManager.gzip` line. 

So, I wanted to add both some logs here to get some more insight in our logs and also add a `timeout` for that gzipping phase so we don't spent too much time there unnecessarily. Tested locally and it works well on my machine.


soulgalore:
Cool thank you @canova !

PR #1902, opened 2023-02-15
Fix visual-complete progress calculations to not use fuzz/slop.

gmierz:
I found a bug in the way the visual-complete progress gets calculated. It's related to the usage of slop in the existing code which shouldn't be done. Here's some more information from a comment I added to the new method:
```
    This method finds the visually-complete progress by taking a sum of
    all the differences in the histogram between the current frame, and the
    final frame. The initial/first frame is used to remove values that are
    consitent between the first, and final frame (i.e. it's a baseline).

    Note that this method should not be using a slop/fuzz because we aren't
    looking at individual pixel intensities which is where the fuzz can be
    found. Within individual channels, we hit an issue where we cannot tell
    if a value in the red channel is purely from a red colour in the image, or
    if it's mixed with other colours such as grey (e.g. red with rbg(240,0,0),
    and light grey with rgb(245,245,245)).
```

You can find more information here: https://bugzilla.mozilla.org/show_bug.cgi?id=1815970#c1
With this change, I was able to resolve an invalid performance regression that we had here: https://bugzilla.mozilla.org/show_bug.cgi?id=1809874

@soulgalore what do you think about this?

### DIFF
--- a/browsertime/visualmetrics-portable.py
+++ b/browsertime/visualmetrics-portable.py
@@ -372,7 +372,7 @@ def video_to_frames(
     viewport_retries,
     viewport_min_height,
     viewport_min_width,
-    full_resolution
+    full_resolution,
 ):
     """Extract the video frames"""
     global client_viewport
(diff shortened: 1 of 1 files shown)


soulgalore:
Hi @gmierz thank you for this! Sorry for the late feedback, my family been sick last week so haven't had any time at the computer. This is really interesting since it's been a problem in long time with finding the "correct" last change (or at least I've also seen cases where it been wrong.

gmierz:
Hi @soulgalore, no problem, I hope they're all getting better! Thanks for merging the fix!

PR #1904, opened 2023-02-17
Ignore browsing context events without requests

juliandescottes:
If no request can be found for a browsing context event, we can't compute the page timings and it will confuse our current HAR generation.


juliandescottes:
While testing the har module on a few more navigations I had an issue with some request-less browsing context creation. I think we should simply ignore those events for now. Once we have page performance timings in those events (https://github.com/w3c/webdriver-bidi/issues/357), we could revert this.

soulgalore:
Ah cool thank you @juliandescottes!

PR #1908, opened 2023-02-24
Use firefox-devtools/bidi-har-export to record BiDi HAR

juliandescottes:
Did a quick test with `node bin/browsertime.js https://www.sitespeed.io --browser firefox` and it successfully generated a HAR file.

For now we don't plan to publish this on npm, which is why I am directly referencing a github commit. But if at any point we want to really use this for browsertime, we can publish a package.


soulgalore:
Great, thank you @juliandescottes, so when will the bidi support be in stable Firefox? I want to do some testing but don't want to start to early :)

juliandescottes:
> Great, thank you @juliandescottes, so when will the bidi support be in stable Firefox? I want to do some testing but don't want to start to early :)

The BiDi events we use are already in Firefox release, so you can use this with the regular release channel of Firefox.
I am still expecting bugs to be found once we start testing this against more exotic scenarios, but if you want to give it a go and test it further, your feedback would be super helpful :)

soulgalore:
Aha, ok, so I could use this directly with the stable channel? Let me make it configurable which version to use then, I can push it to the test I continuously run and use the new version there.

Issue #1912, opened 2023-03-14
Geckoprofiler ending too early


92kns:
### Your question

Hi @soulgalore ! 
While running the gecko profiler on benchmark tests via Browsertime (e.g. a browser benchmarks such as [ARES-6](https://browserbench.org/ARES-6/index.html), [speedometer2](https://browserbench.org/Speedometer2.0/), [MotionMark](https://browserbench.org/MotionMark/)) , we find that gecko profiler ends too early or sometimes before the benchmark has even started.

I suspect the `afterPageCompleteCheck`: https://github.com/sitespeedio/browsertime/blob/081d93e1aeb6805f4fb96df07815d4a26919676e/lib/firefox/webdriver/firefox.js#L111-L122 is being triggered earlier than desired. 

Would you have any ideas/thoughts on if there a way to control the length of the geckoprofiler run so we could delay/control the `geckoProfiler.stop()`? I looked at the Browsertime options but doesn't seem like there is any that would specifically do that
Thanks!

soulgalore:
Hi @92kns when you do need the profiler to end? This is how the page complete check works: https://www.sitespeed.io/documentation/sitespeed.io/browsers/#choose-when-to-end-your-test - the idea is that when that signals that the test is over, we shutdown gecko profiler and the same functionality for Chrome, then we start to run the JavaScript metrics (we run JS to collect metrics from the browser). So either if you can define when you want to end the test or if I you need collect data after the test, we need to add some kind configuration/switch for that.

92kns:
Thanks for the speedy response and for the link, that gives me a better idea of the sequence of events.
> when you do need the profiler to end?

it varies depending on the benchmark but I think having the option to keep the profiler running for even an extra 5-10 seconds would be a good start. So I tried playing around with `--pageCompleteWaitTime` and that seemed to increase the profiler run time! (but made other things slow - I need to play around with it a bit more 😅 )

inadvertently, this also helped me find an issue with another benchmark (it seems to consistently only trigger after the gecko profiler stops, no matter how long I increased `--pageCompleteWaitTime`)

Anyway, thanks for your help. I will let you know as soon as I can if we've resolved the issue and can close this ticket, or if I need some more help!

92kns:
Hi @soulgalore we were thinking of exposing the geckoprofiler through the commands so that we might be manually control the .start()/.stop() ourselves

[...]

PR #1934, opened 2023-04-05
Expose geckoprofiler start and stop to commands 

92kns:
For issue #1912

This lets us expose the start/stop commands for the geckoProfiler object in browsertime, letting us manually choose when to start & stop profiling.


92kns:
i think [3ca680a](https://github.com/sitespeedio/browsertime/pull/1934/commits/3ca680af47c0199e9b828b7e851e7ac138c48ab8) might fix some of the errors... before I was checking the browser and the custom flag at the _same time_ in iterations.js, but since the new flag is part of the firefox group as per the cli.js, then geckoProfilerCustom would be undefined rather than boolean for non firefox browsers I think? So now the check happens _after_ checking if the browser is firefox.

Or maybe I should putting geckoProfilerCustom in a different group since it is being used in iteration.js?

lib/core/engine/iteration.js
```diff
@@ -147,13 +148,32 @@ export class Iteration {
         options
       );
 
+      let profiler = {};
+      if (browser.options.browser == 'firefox') {
```
soulgalore, line 152:
```suggestion
      if (options.browser === 'firefox') {
```

lib/core/engine/iteration.js
```diff
@@ -147,13 +148,32 @@ export class Iteration {
...
+        if (this.options.firefox.geckoProfilerCustom) {
```
soulgalore, line 153:
```suggestion
        if (options.firefox.geckoProfilerCustom) {
```

lib/firefox/geckoProfiler.js
```diff
@@ -121,6 +121,14 @@ export class GeckoProfiler {
     const runner = this.runner;
     const storageManager = this.storageManager;
     const options = this.options;
+    // automatically set the index and url when using the exposed
+    // geckoprofiler commands
+    if (index == undefined && this.firefoxConfig.geckoProfilerCustom) {
+      index = 1;
+    }
+    if (url == undefined && this.firefoxConfig.geckoProfilerCustom) {
+      url = options.browsertime.url;
```
soulgalore, line 130: I'm not sure this will work because sometimes we get the URL direct from the script file and we can also get it after a navigation has happened (clicking on a link etc)?

soulgalore (commented):
There's a couple of fixes I think that needs to be done to get it to work. I'm not sure about the URL and the index, need to try it.

lib/core/engine/iteration.js
```diff
@@ -147,13 +148,32 @@ export class Iteration {
         options
       );
 
+      let profiler = {};
+      if (options.browser === 'firefox') {
+        if (options.firefox.geckoProfilerCustom) {
```
soulgalore, line 153:
```suggestion
        if (options.firefox && options.firefox.geckoProfilerCustom) {
```

soulgalore (commented):
Some small fixes.

lib/support/cli.js
```diff
@@ -466,6 +466,12 @@ export function parseCommandLine() {
       type: 'boolean',
       group: 'firefox'
     })
+    .option('firefox.geckoProfilerCustom', {
+      describe: 'Expose the start/stop commands for the gecko profiler',
+      default: false,
+      type: 'boolean',
+      group: 'firefox'
+    })
```
canova, line 474: `firefox.geckoProfilerCustom` name is a bit vague to me. How about making it something like `geckoProfilerRecordingType`, which can take an enum of either 'pageload' or 'custom'. and default it to 'pagelod'?

canova (commented):
Thanks @92kns! I think it will be very valuable to have start and stop commands available in browsertime!

lib/support/cli.js
92kns: Good idea, perhaps part of firefox.geckoProfilerParams? an optional cli flag to pass like `firefox.geckoProfilerParams.recordingType`

92kns: (or even firefox.geckoProfiler.recordingType)

canova: I think `firefox.geckoProfilerParams` is different. The options under this scope are the parameters we pass inside the profiler start function as arguments. So I think it's better to separate this with them.

I like the `firefox.geckoProfiler.recordingType` idea! That way, it's clear that it's related to the `firefox.geckoProfile` as it's scoped that way.

92kns:
I think the code looks better making this change now from geckoProfilerCustom -> geckoProfiler.recordingType and also should prevent a user from (accidentally?) using both flags (which could have happened in the previous implementation)

lib/support/cli.js
```diff
@@ -466,6 +466,13 @@ export function parseCommandLine() {
       type: 'boolean',
       group: 'firefox'
     })
+    .option('firefox.geckoProfiler.recordingType', {
+      describe: 'Expose the start/stop commands for the gecko profiler',
+      default: 'pageload',
+      choices: ['pageload', 'custom'],
+      type: 'string',
+      group: 'firefox'
+    })
```
92kns, line 475: Hm does this implementation make sense?

92kns:
My mistake, realize this will follow geckoProfiler path regardless if i use custom start/stop. I will try to fix this shortly

lib/firefox/geckoProfiler.js
soulgalore: Sorry, I mean the start/stop command looks cool, however if it will work IRL is another story :D 

I think internally the stop command needs to be able to get the correct URL and correct index, so that the geckoprofilelog will be stored at the correct place. I think you could follow the same pattern as with the screenshot command https://github.com/sitespeedio/browsertime/blob/main/lib/core/engine/command/screenshot.js - that class needs the iteration and can run JavaScript (so you can collect the current URL from the browser), Does that make sense?

92kns: Thanks for the link, so I could use 
```javascript
const url = await this.browser
      .getDriver()
      .executeScript('return document.documentURI;');
```
to get the url.
Is there a similar way to auto grab the index?
`this.browser.options.index` I noticed this field, could I use this?

lib/firefox/geckoProfiler.js
```diff
@@ -121,6 +121,22 @@ export class GeckoProfiler {
...
+      index = runner.options.index + 1;
+    }
+    if (
+      url == undefined &&
+      this.firefoxConfig.geckoProfilerRecordingType === 'custom'
+    ) {
+      url = await runner
+        .getDriver()
+        .executeScript('return document.documentURI;');
+    }
```
92kns, line 139: how is this @soulgalore ?

lib/firefox/geckoProfiler.js
soulgalore: I think you can follow the example exactly by making a new GeckoprofileCommand (or something), then when it's instantiated in the iterations, it can get the index the same ways as the screenshot class: https://github.com/sitespeedio/browsertime/blob/main/lib/core/engine/iteration.js#L176 

Then the class can have start/stop so the user more easily can handle it.

I can help tomorrow.

92kns: Okay that makes sense now thank you!
The new push has an attempt at this implementation

soulgalore:
Looks good @92kns thank you! I will merge this tonight and do a follow up PR with some small small fixes.

92kns:
Awesome, thanks for your help @soulgalore !

PR #1936, opened 2023-04-07
Firefox HAR 0.0.7

soulgalore:
New version with fixes + possibility to enable logging.


soulgalore:
Thank you @juliandescottes for the new version!

Issue #1958, opened 2023-06-16
Debug mode is currently broken for desktop


gmierz:
### Have you read the documentation?

- [X] Yes, I've read the [how to make a reproducable bug guide](https://www.sitespeed.io/documentation/sitespeed.io/bug-report/)
- [X] Yes, I've read the [how to debug my script guide](https://www.sitespeed.io/documentation/sitespeed.io/scripting/#debug)

### URL

Any

### What are you trying to accomplish

When running browsertime with --debug mode, we get this error message on desktop: `Debug mode do not work on Android. Please run debug mode on Desktop.`

### What browser did you use?

Firefox

### How to reproduce

```shell
Run this command: `browsertime/bin/browsertime.js https://www.sitespeed.io --browser firefox --debug`
```

gmierz:
This is fixed now with the merged PR.

PR #1959, opened 2023-06-16
Issue #1958 - Check for android.enabled when checking debug mode.

gmierz:
This patch fixes an issue with --debug mode on desktop (see issue). The hack for backwards compatibility is moved into the verification part of arg parsing code instead of after it.

### DIFF
--- a/lib/support/cli.js
+++ b/lib/support/cli.js
@@ -63,6 +63,11 @@ function validateInput(argv) {
     }
   }
 
+  // hack to keep backward compability to --android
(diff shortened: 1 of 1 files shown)


soulgalore:
Thank you @gmierz ! I wonder if there's a smart way to test this in a GitHub Action so it doesn't happen again?

gmierz:
+1 I think we would need a test that checks that the error isn't thrown with a few combinations of commands

Issue #1968, opened 2023-07-19
Chromedriver 115 do not find Chrome


soulgalore:
When I updated to 115.0.5790-98 I get `cannot find Chrome binary` even though I got Chrome 115 installed on my Mac. I wonder if there's a new switch or something and that Chromedriver defaults to "Chrome for testing"? It seems to work fine for Linux and Windows though. On Mac it works if you set the path to the Chrome binary.

soulgalore:
https://github.com/GoogleChromeLabs/chrome-for-testing/issues/30 is probably the root cause. Let's wait on that before re-releasing 115 again.

PR #1981, opened 2023-08-15
New Chromedriver version that fixes the Mac issue.

soulgalore:
Fixes
https://github.com/sitespeedio/browsertime/issues/1968 so we can renmove our fix.


soulgalore:
We don't need this with the new 116

PR #1999, opened 2023-10-12
Get the url from the `result` array in GeckoProfiler.stop

canova:
Previously for gecko profiler commands we were getting the test url by running a script that returns `document.documentURI`. This was working fine for most cases. But when one of our tests actually navigates to a different page from the starting page, it gets the destination url instead of the initial url. Since we need the initial url, it's better not to get the url like this, and get it directly from the `result` array.

Note that chrome trace command does something similar [here](https://github.com/sitespeedio/browsertime/blob/e3e5a46f6c9070c0ab66e543b85e60d021edc848/lib/core/engine/command/chromeTrace.js#L42).

cc @soulgalore and @92kns 
This fixes [Bug 1858709](https://bugzilla.mozilla.org/show_bug.cgi?id=1858709)

### DIFF
--- a/lib/core/engine/command/geckoProfiler.js
+++ b/lib/core/engine/command/geckoProfiler.js
@@ -1,11 +1,12 @@
 import intel from 'intel';
 const log = intel.getLogger('browsertime.command.geckoprofiler');
 export class GeckoProfiler {
-  constructor(GeckoProfiler, browser, index, options) {
+  constructor(GeckoProfiler, browser, index, options, result) {
     this.GeckoProfiler = GeckoProfiler;
     this.browser = browser;
     this.index = index;
     this.options = options;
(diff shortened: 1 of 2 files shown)


soulgalore:
Cool, let me rerun the tests to see the unit tests come through.  There are two other actions that will fail at the moment: Chromedriver/Chrome has some breaking changes in the upcoming 119 release (the current driver doesn't work with upcoming releases) so lets merge with those failing.

canova:
@soulgalore thanks! It looks like the unit tests are failing again but they are passing locally.

soulgalore:
Yeah, the tests always work on my machine too ;) 

Something is flakey in that action I will have a look at it later on. Thanks for the PR @canova

PR #2018, opened 2023-11-13
Gather LargestContentfulPaint in Firefox.

gmierz:
This patch adds the ability to gather LargestContentfulPaint with Firefox. It's gathered into a `firefoxWebVitals` entry similar to what is done with Chrome-based browsers.

### DIFF
--- a/lib/core/engine/collector.js
+++ b/lib/core/engine/collector.js
@@ -47,7 +47,6 @@ function getNewResult(url, options) {
     deltaToTTFB: [],
     cpu: [],
     googleWebVitals: [],
-    firefoxWebVitals: [],
     extras: [],
     fullyLoaded: [],
     mainDocumentTimings: [],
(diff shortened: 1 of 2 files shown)


gmierz:
Hi @soulgalore! Let me know what you think about this approach. I was thinking it might be best to keep these in a single entry (maybe called `webVitals`), but it looks like the `googleWebVitals` is used quite a bit everywhere so I didn't change anything there.

soulgalore:
Hi @gmierz let me get back on that later today or early tomorrow. I think the metrics should work and not need to be added for Firefox, but let me check. I tried the LCP in Firefox nightly sometime ago and it worked fine. I also compared it with the largest contentful paint for the video and it most cases it correlated.

Let me just explain why it's named Google Web Vitals: So ... Google comes up with a series of metrics they say are vital to the web. I don't think they are in the position to define what's vital all across the web. What's vital depends on the web site and the user. But I guess these metrics are vital for getting hight score in Google SEO, so that's why its names Google Web Vitals.

soulgalore:
@gmierz what was the idea with this PR? :) I see that we don't add Firefox metrics to the GoogleWebVitals field (missing FCP and LCP), maybe that is the fix that needs to be done? The standalone metric is collected by default, so if I run:
`browsertime -b firefox --firefox.nightly https://www.sitespeed.io` the largest contentful paint is there.

gmierzwinski:
Ah ok, that explains it, thanks! :)

I wasn't sure how we should handle the metric from Firefox. I'll change this to add it to `googleWebVitals` instead.

gmierz:
Ok, thanks for the review @soulgalore, I've removed the `firefoxWebVitals` entry and added it to the `googleWebVitals` instead.

soulgalore:
Thank you @gmierz !

PR #2027, opened 2023-12-09
Update dev dependencies

soulgalore:


soulgalore:
Oops that auto-update didn't go the way I wanted :)

Issue #2041, opened 2023-12-20
Firefox 121 on Mac OS (14.2)  fails to start


soulgalore:
It's the same on M1 and old AMD.
```
bin/browsertime.js https://wwww.sitespeed.io -b firefox -n 1
[2023-12-20 04:55:26] INFO: Running tests using Firefox - 1 iteration(s)
[2023-12-20 04:55:27] INFO: firefox failed to start, trying 2 more time(s): Process unexpectedly closed with status signal
[2023-12-20 04:55:27] INFO: firefox failed to start, trying 1 more time(s): Process unexpectedly closed with status signal
[2023-12-20 04:55:27] INFO: firefox failed to start, trying 0 more time(s): Process unexpectedly closed with status signal
```

Running on Ubuntu seems to work fine.

From Geckodriver:
```
1703046910184	geckodriver	INFO	Listening on 127.0.0.1:52620
1703046910195	webdriver::server	DEBUG	-> GET /status 
1703046910195	webdriver::server	DEBUG	<- 200 OK {"value":{"message":"","ready":true}}
1703046910209	webdriver::server	DEBUG	-> POST /session {"capabilities":{"firstMatch":[{}],"alwaysMatch":{"browserName":"firefox","pageLoadStrategy":"none","moz:fir ... ry":"/Applications/Firefox.app/Contents/MacOS/firefox-bin","args":["-no-remote","-devtools"],"env":{}},"webSocketUrl":true}}}
1703046910211	geckodriver::capabilities	DEBUG	Trying to read firefox version from ini files
1703046910213	geckodriver::capabilities	DEBUG	Found version 121.0
1703046910214	geckodriver::capabilities	DEBUG	Creating profile directory tree /var/folders/jx/m9wvcdl12kx2dt9dznxjt3mr0000gn/T/rust_mozprofile5FLLGU/chrome/
1703046910214	geckodriver::capabilities	DEBUG	Creating profile directory tree /var/folders/jx/m9wvcdl12kx2dt9dznxjt3mr0000gn/T/rust_mozprofile5FLLGU/extensions/
1703046910214	geckodriver::capabilities	DEBUG	Extracting profile to /var/folders/jx/m9wvcdl12kx2dt9dznxjt3mr0000gn/T/rust_mozprofile5FLLGU/chrome/userChrome.css
[...]

PR #2043, opened 2023-12-21
Fix Firefox 121 start issue on Mac.

soulgalore:
See https://github.com/sitespeedio/browsertime/issues/2041

### DIFF
--- a/lib/firefox/webdriver/builder.js
+++ b/lib/firefox/webdriver/builder.js
@@ -1,6 +1,7 @@
 import { resolve } from 'node:path';
 import { fileURLToPath } from 'node:url';
 import path from 'node:path';
+import { platform } from 'node:os';
 import {
   ServiceBuilder,
   Options,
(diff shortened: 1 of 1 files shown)


lib/firefox/webdriver/builder.js
```diff
@@ -168,9 +169,36 @@ export async function configureBuilder(builder, baseDir, options) {
...
-    );
+    // Fix for https://github.com/sitespeedio/browsertime/issues/2041
+    if (platform() === 'darwin') {
+      if (
+        !firefoxConfig.binaryPath &&
+        !firefoxConfig.nightly &&
+        !firefoxConfig.beta &&
+        !firefoxConfig.developer
+      ) {
+        ffOptions.setBinary('/Applications/Firefox.app/Contents/MacOS/firefox');
```
whimboo, line 180: @soulgalore actually it would be better if you only set the app bundle on MacOS. This would prevent issues with the real binary that you usually should read out of the `info.plist` file. geckodriver itself will find the binary on its own.

soulgalore: I'm gonna remove this fix when it's fixed upstream. It was bold releasing 121 the week before Christmas so I just want to get it to work :)

whimboo: Ah, that's fine then. Thanks!

diemol: It would be better to let Selenium Manager handle this. It will find the Firefox binary on the host and then match it with the appropriate GeckoDriver version.

PR #2052, opened 2024-01-03
Remove the default Firefox profile.

soulgalore:
That profile was created 7 years ago. I think the idea was to have certs working for WebPageReplay and there is also some CSS fix for SpeedIndex. I think those fixes are long overdue so lets remove them.


soulgalore:
Want to get this in the release going out this weekend.

soulgalore:
Np thank you @gmierz :)

PR #2087, opened 2024-03-06
Report some Loaf data in upcoming Chrome

soulgalore:
Lets iterate and see which metrics is really valuable to fetch.


soulgalore:
Lets start with this and then we can iterate.

PR #2090, opened 2024-03-08
Cookie using Bidi for Firefox 124

soulgalore:
This works in Nightly, not sure if it comes in 124 or 125.


soulgalore:
This works in 124.

PR #2108, opened 2024-03-19
Preparing for setting headers when FF supports it in Bidi (Firefox 1XX)

soulgalore:
> '"headers" not supported yet in network.continueRequest',

Example on how to set request headers in Bidi if will be available in continueRequest?


lib/firefox/firefoxBidi.js
```diff
@@ -146,4 +147,54 @@ export class FirefoxBidi {
...
+    await bidi.send(command);
+    await bidi.subscribe('network.beforeRequestSent');
+    await bidi.onMessage(async function (event) {
+      const parsedEvent = JSON.parse(Buffer.from(event.toString()));
+      if (parsedEvent.method === 'network.beforeRequestSent') {
+        const continueRequest = {
+          method: 'network.continueRequest',
+          params: {
+            request: parsedEvent.params.request.request,
+            headers
```
juliandescottes, line 188: Mentioned it on bugzilla, but here you should merge your headers with the current request headers.
```suggestion
            [headers](headers: [...parsedEvent.params.request.headers, ...headers],)
```

soulgalore: Ah thank you! I got confused since I got "success" without the rest of the headers.

Issue #2114, opened 2024-03-20
Safari Technology Preview not being recognized as a browser?


92kns:
### Your question

It is possible I am doing something wrong.

When using the following flags `--browser safari` and `--safari.useTechnologyPreview` I get the following error

```
 safari failed to start, trying 2 more time(s): Could not create a session: Browser name does not match (requested: safari; available: Safari Technology Preview
```

Possible bug? Safari-tp should be part of the Safari group right?

The driver isn't being configured here it seems https://github.com/sitespeedio/browsertime/blob/d17537bdffb6ca9e18e3933978b4013e9650307b/lib/core/seleniumRunner.js#L84-L89

which goes here https://github.com/sitespeedio/browsertime/blob/d17537bdffb6ca9e18e3933978b4013e9650307b/lib/core/webdriver/index.js#L43-L46

but this file seems fine https://github.com/sitespeedio/browsertime/blob/d17537bdffb6ca9e18e3933978b4013e9650307b/lib/safari/webdriver/builder.js since `builder` seems to return just fine. So I am a bit unsure why it is failing.

it is worth mentioning that I have ran `safaridriver --enable` with the safaridriver that comes bundled within safari technology preview, and if I manually open safari-tp myself, the automation field in the developer setting is all checked.

any ideas what I might be doing wrong?

soulgalore:
I think maybe this is a bug in the NodeJS flavour of Selenium but I need to make a test case to test it. Looking at the Selenium code for Python that handles the browser name "Safari Technology Preview" but I couldn't find the same in NodeJS.

92kns:
@soulgalore thanks for taking a look!

That error also threw me off.

I did some hacky stuff locally and it *kind of* works?

 So in the NodeJS part of selenium here https://github.com/SeleniumHQ/selenium/blob/12344fe74af55ee6731ccda860b63fd77e6ffd96/javascript/node/selenium-webdriver/safari.js#L65
 It seems `safari` is hard coded here, when it should be `Safari Technology Preview` when running safari-tp

and here there is now option for that https://github.com/SeleniumHQ/selenium/blob/12344fe74af55ee6731ccda860b63fd77e6ffd96/javascript/node/selenium-webdriver/lib/capabilities.js#L31-L37

[...]

PR #2115, opened 2024-03-21
Set browser name to Safari TP for selenium options

92kns:
For #2114

This seems to work for me now. Wouldn't mind a double check 😄


soulgalore:
Thanks @92kns that made it work!

PR #2119, opened 2024-05-02
Add support to visualmetrics to identify key frames matching the given colors

aosmond:
This patch updates the visualmetrics.py and visualmetrics-portable.py scripts to allow it to identify and record timestamps for frames that match the given color configuration. This consists of an RGB value with fuzz (anything +/- fuzz matches) and a fraction between 0 and 1 of the required percentage of each channel from the histogram that must match it.

This is intended to be used in Firefox automation to collect metrics with respect to video playback latency.


gmierz:
Hi @aosmond, interesting metric! Can you provide an example video, and command that we could test this with along with the expected output from it?

aosmond:
I added an example zip/description here:
https://gist.github.com/aosmond/4daa941cf36670e2ab0f7ecbeb66ed00

This should yield in browsertime.json under "visualmetrics" when the poster, first video frame, second video frame and last video frame are displayed.

visualmetrics/visualmetrics-portable.py
```diff
@@ -1346,6 +1371,90 @@ def load_histograms(histograms_file, start, end):
...
+    key_color_frames = {}
+    for key in key_colors:
+        key_color_frames[key] = []
+
+    current = None
+    current_key = None
+    total = 0
+    matched = 0
+    channels = ["r", "g", "b"]
+    for index, histogram in enumerate(histograms):
```
gmierz, line 1387: Is there any specific reason you chose to use a `for` loop here? The matching/current_key seems a bit over-complicated to me, and it could be simplified by using a couple while loops instead, here's some pseudo-code of what I'm thinking about:
```
        while histograms left:
            check if current histogram matches a key
            if not, continue to next histogram

            while we haven't found the final histogram for this key:
                check if current histogram matches
                    add to framecount
                if not, break out

            set the end time of the key
```

visualmetrics/visualmetrics.py
```diff
@@ -2200,6 +2305,25 @@ def main():
...
+            "green_high",
+            "blue_low",
+            "blue_high",
+            "fraction",
+        ),
+        help="Identify frames that match the given channel (0-255) low and "
+        "high. Fraction is the percentage of the pixels per channel that "
+        "must be in the given range (0-1).",
+    )
+    parser.add_argument("--keycolors", help="Key color frames output file.")
```
gmierz, line 2326: I'm not seeing this being used in the browsertime call for visual metrics - can you elaborate on what the purpose of this argument is?

gmierz (approved):
r+ for the visual metrics script changes, just a nit and a small question below. Thanks for fixing up the calculation method!

lib/support/har/index.js
```diff
@@ -70,7 +70,8 @@ function addExtrasToHAR(
       'ContentfulSpeedIndex',
       'VisualProgress',
       'ContentfulSpeedIndexProgress',
-      'PerceptualSpeedIndexProgress'
+      'PerceptualSpeedIndexProgress',
+      'KeyColorFrames'
```
soulgalore, line 74: This is metrics added to the HAR file so HAR viewers can draw the metrics in the HAR waterfall, would the keycolorframes and any value here?

lib/video/postprocessing/finetune/getTimingMetrics.js
```diff
@@ -52,6 +65,28 @@ export function getTimingMetrics(videoMetrics, timingMetrics, options) {
...
       }
     );
+
+    if (vm.KeyColorFrames) {
+      let keyColorFrames = jsonifyKeyColorFrames(vm.KeyColorFrames);
+      for (let key in keyColorFrames) {
+        for (let frame of keyColorFrames[key]) {
+          let start = frame.startTimestamp;
+          let end = frame.endTimestamp;
+          if (end > start) {
```
soulgalore, line 75: You can fix probably fix most of those errors by running `npm run lint:fix` and then push the changes.

soulgalore (commented):
I'm wondering could this just be a change in visualmetrics-portable.py and then add a generic way to pass on parameters to that script from Browsertime, and then you could have your magic analysing it in your ci?

lib/support/har/index.js
aosmond: It could be. I'm not familiar with the tool and how I would need to massage the data to be compatible with the viewer, so I just excluded it.

lib/video/postprocessing/finetune/getTimingMetrics.js
```diff
@@ -24,17 +25,29 @@ function get(metric, metricName, pos, options) {
     x = 10;
   }
 
+  // Adjust the range by 1ms to ensure it displays on the correct frames.
```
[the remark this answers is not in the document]
aosmond: The default of 30 seems adequate. I actually display the first and second frames a few extra times in the video file being displayed itself to ensure we are more likely to capture. We get it most of the time. But if you have a better alternative, I'd love to hear it. The Firefox window recorder always captures everything quite nicely for example, but that doesn't map to Chrome/Safari very well.

aosmond:
> I'm wondering could this just be a change in visualmetrics-portable.py and then add a generic way to pass on parameters to that script from Browsertime, and then you could have your magic analysing it in your ci?

I'm happy to make whatever changes are necessary for a merge :). But I'm a bit uncertain as to what that would look like, as I didn't make too many changes outside the python scripts, just what I thought was necessary to get the information out and annotate the videos. What parts would you like removed?

I already do a post processing phase on the raw key color data in our CI to normalize things in case it didn't get a first frame, but found a second frame, etc.

visualmetrics/visualmetrics.py
gmierz: Ah ok, you could leave it if you want, but thanks for removing it!

lib/support/har/index.js
soulgalore: Sorry what I meant was I think you can safely remove it from there, thanks :)

lib/video/postprocessing/finetune/getTimingMetrics.js
```diff
@@ -24,17 +25,20 @@ function get(metric, metricName, pos, options) {
...
+  let range = Number.isFinite(end)
+    ? `between(t,${(Number(start) - 1) / 1000},${(Number(end) - 1) / 1000})`
+    : `gte(t,${(Number(start) - 1) / 1000})`;
   // H-h/8
-  return `drawtext=${fontFile}enable='between(t,${
-    Number(metric) / 1000
-  },30)':x=(w-tw)/2: y=H-${pos}-h/${x}:fontcolor=white:fontsize=h/${fontSize}:box=1:boxcolor=0x000000AA:boxborderw=2:text='${metricName} ${metric}'`;
+  return `drawtext=${fontFile}enable='${range}':x=(w-tw)/2: y=H-${pos}-h/${x}:fontcolor=white:fontsize=h/${fontSize}:box=1:boxcolor=0x000000AA:boxborderw=2:text='${metricName} ${start}'`;
```
soulgalore, line 33: Do you need to display it on the video or is just getting the metrics in the JSON enough? The reason I ask is that the more changes done in the original code, the more I need to test and verify that old functionality still works.

lib/support/cli.js
```diff
@@ -718,6 +723,12 @@ export function parseCommandLine() {
       describe:
         'Use the portable visual-metrics processing script (no ImageMagick dependencies).'
     })
+    .option('visualMetricsKeyColor', {
+      type: 'array',
+      nargs: 8,
```
soulgalore, line 728: Think this is missmatch to the gist you provide with how to run it? I tried with `bin/browsertime.js --visualMetrics true --video true --iterations 1 --visualMetricsKeyColor poster 120 250 250 20 0.75 --visualMetricsKeyColor first_frame 250 20 20 20 0.75 --visualMetricsKeyColor second_frame 20 20 250 20 0.75 --visualMetricsKeyColor last_frame 255 255 50 50 0.75 --browser firefox --video.framerate 60 https://www.sitespeed.io` (I think that is in the example) but then I get "Not enough arguments following: visualMetricsKeyColor". Can you share how I can try it + do you know a live site where I can try?

aosmond: I updated my gist at https://gist.github.com/aosmond/4daa941cf36670e2ab0f7ecbeb66ed00. I don't have a live site. It's primary purpose is to measure the latency of video playback, hence the special rainbow video I made, but in theory any video that displays a non-white solid color frame at some point would work.

lib/video/postprocessing/finetune/getTimingMetrics.js
aosmond: It would be nice to put in the review, but I removed it to make your review simpler. Maybe I can provide it in a future PR of a lower priority :).

PR #2134, opened 2024-05-21
Add the ability to gather power usage measurements on Android from USB power meters.

gmierz:
This PR adds the ability to gather power from USB power meters that are supported by the `usb-power-profiling` module: https://github.com/fqueze/usb-power-profiling

The capability is added to both Firefox, and Chromium-based browsers. Some methods are added to the Android class to make it simpler to parse the power usage measurements, and also obtain a full "power profile" of the power usage throughout the full test. The power profile is gathered for each test-iteration/url measured.


gmierz:
Hi @soulgalore! What do you think about adding the `usb-power-profiling` module for power measurements on android using USB power meters?

We've found that this is a better approach versus the batterystats method that currently exists in the Android class. If you want, I could remove that older power testing code and make this the primary method of doing power usage testing in Browsertime. The batterystats method is also device-specific whereas this method only requires a supported USB power meter to be able to test on any Android phone.

soulgalore:
Hi @gmierz sounds like a good idea, let me test it tomorrow!

soulgalore:
Hmm the unit tests started to fail, do they work for you locally @gmierz ?

gmierz:
That's an odd failure. I can reproduce it locally though so I'll look into it.

gmierz:
It works now! There was some code in the usb-power-profiling module that would get run on import and this caused the timeout. I've got a PR up on usb-power-profiling to fix it: https://github.com/fqueze/usb-power-profiling/pull/4

Once it's merged, I'll update the the package used in this PR (currently it's using a custom git commit).

soulgalore:
Looks good, lets merge when your fix is published.

soulgalore (approved):

PR #2176, opened 2024-09-03
Call stopSampling on browser stop during android power testing.

gmierz:
This patch resolves an issue that occurs on some systems where the power testing needs to explicitly be stopped before the next iteration.


soulgalore:
Thank you @gmierz !

gmierz:
Thanks for the review, and merge @soulgalore :)

PR #2220, opened 2024-12-09
Improve how we compute the power consumption numbers from the Firefox Profiler output

canova:
Hey @soulgalore!

I was looking at some of the profiles and the powerConsumption metrics that's generated using the Firefox Profiler. And realized that there were some issues related to how we compute this number.

This feature was first implemented [here](https://github.com/sitespeedio/browsertime/pull/2046) (and [this is the issue](https://github.com/sitespeedio/sitespeed.io/issues/3944))

1. We were only looking at only the parent process counters. Because child process profiles are stored in the parent process' `processes` field, so now we look at the profiles recursively and sum all the power consumption counters coming from all the processes.
2. I excluded some of the power counters, because sometimes we were counting the power usage values twice. See the message of this commit for more details: https://github.com/sitespeedio/browsertime/commit/3003c71d50a155bd1ef7cf700d9b289a0f77f378
3. I change the counter loop to exclude the first value of the `count` values. In some platforms the first value can either be different, and might mean other things, or it can be incorrect. [We already have this logic in the profiler frontend](https://github.com/firefox-devtools/profiler/blob/6a683fa6ae2733dae49ff460edf6ee98ca626f09/src/profile-logic/profile-data.js#L1740-L1743) to ignore the first value for all the counters, so we weren't hitting this issue before. But I encountered this issue while I was comparing the values from the profiler frontend vs the browsertime.

I recommend looking at the PR commit-by-commit as the commit messages also provide some context.

To be able to test locally, you can follow the instructions here: https://github.com/sitespeedio/sitespeed.io/issues/3944#issuecomment-1874409458

### DIFF
--- a/lib/firefox/geckoProfiler.js
+++ b/lib/firefox/geckoProfiler.js
@@ -33,7 +33,20 @@ function computeCounterPowerSum(counter) {
 function computeFullProfilePower(profile) {
   let power = 0;
   for (const counter of profile.counters) {
-    if (counter.category === 'power') {
+    if (
+      counter.category === 'power' &&
+      // If power counters starts with "Power:", it means that there are several
+      // individual components that Firefox records. For Linux and Windows,
+      // intel CPUs might produce 'CPU package' to show the whole power
+      // consumption of the CPU. But on top of that it also shows some individual
+      // components like, DRAM, iGPU, individual CPU cores etc. We don't want
+      // to include them as it might produce double the power values it normally consumes.
+      //
+      // Linux track names: https://searchfox.org/mozilla-central/rev/5ad94327fc03d0e5bc6aa81c0d7d113c2dccf947/tools/profiler/core/PowerCounters-linux.cpp#48-70
+      // Windows track names: https://searchfox.org/mozilla-central/rev/5ad94327fc03d0e5bc6aa81c0d7d113c2dccf947/tools/profiler/core/PowerCounters-win.cpp#32-55
+      (!counter.name.startsWith('Power:') ||
+        counter.name === 'Power: CPU package')
+    ) {
       power += computeCounterPowerSum(counter);
     }
   }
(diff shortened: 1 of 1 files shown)


soulgalore:
Hi @canova thanks for the fixes, I'll take a look tonight. So there's another thing about the calculations in https://github.com/sitespeedio/sitespeed.io/issues/4329 - I've verified that too: if you test multiple URLs within one run, the cpu metrics in the trace seems to increase. Is that how it works (e.g we need to change our implementation on our side if a user tests multiple pages).

canova:
@soulgalore Ah, I wasn't aware of this issue.

It seems like it's because of the 3rd item I mentioned in the list above, which is being fixed here: https://github.com/sitespeedio/browsertime/pull/2220/commits/9504fc78a13f9e8f6095f6fc6d44af752154d8ff

To give more context, it seems like the first values in the counter samples have different meanings per platform. I'm assuming you and @jannekalliola have macOS with Apple Silicon chips, because this is the behavior I encountered while I was testing too.

It seems like when we call the macOS API for getting the power usage values for the first time, it's returning a very high number, which @fqueze suspected that it's the power value since the process start time. This explains why we are seeing increasing amount of values in the sitespeed.io power metrics. It's because it's never being reset, so it's always include the power usage numbers since the process has started.

To fix this, we simply ignore the initial values now in https://github.com/sitespeedio/browsertime/pull/2220/commits/9504fc78a13f9e8f6095f6fc6d44af752154d8ff.

I tested with this PR applied on sitespeed.io and it seems like we are now getting expected values, like:
0.0036 µWh
0.0019 µWh
0.0020 µWh

Also, as I side note, the units in the sitespeed.io results look incorrect. They should be `Wh` instead of `µWh`. µWh is a lot smaller. We have this function to choose the best unit depending on the number: https://github.com/firefox-devtools/profiler/blob/6306555856f9c5b1f539462f24c182b7c07d477b/src/components/tooltip/TrackPower.js#L84-L115 (it's not obvious but `l10nId` is the unit we choose.)
But I can file a different bug for this one.

PR #2275, opened 2025-04-04
Chrome/driver 135, Edge 135 & Firefox 137

soulgalore:


soulgalore:
Update to latest instead.

PR #2279, opened 2025-05-06
Add commands to start/stop and collect PerfStats performance counters

dpalmeiro:
This PR adds some new commands to control PerfStats in the Firefox browser.

### DIFF
--- a/lib/core/engine/command/perfStats.js
+++ b/lib/core/engine/command/perfStats.js
@@ -1,0 +1,67 @@
+import { PerfStats } from '../../../firefox/perfStats.js';
+/**
+ * Manages the PerfStats interface used for collecting Firefox performance counters.
+ *
+ * @class
+ * @hideconstructor
+ */
+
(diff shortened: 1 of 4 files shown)


soulgalore:
Hi @dpalmeiro what's up with that binary file?

dpalmeiro:
Oops, thanks for catching that.  I've removed it now.

soulgalore:
Thank you @dpalmeiro ! Is there any documentation about the perf counters or is it more an internally Firefox thing?

dpalmeiro:
> Thank you @dpalmeiro ! Is there any documentation about the perf counters or is it more an internally Firefox thing?

Yes, there is documentation about this available here:  https://firefox-source-docs.mozilla.org/performance/perfstats.html
It's not widely used, but does give us access to some low overhead specific measurements such as the amount of time spent performing layout, gc, etc.

soulgalore:
Cool thank you @dpalmeiro

PR #2297, opened 2025-05-30
fix: modernize logger interface

emmanuel-ferdman:
## PR Summary
The `logger.warn()` method is deprecated since Python2.7 and replaced with `logger.warning()`. It leads to those warnings:
```python
DeprecationWarning: The 'warn' method is deprecated, use 'warning' instead
```
It also fixes a few typos along the way.


soulgalore:
Thank you @emmanuel-ferdman

PR #2301, opened 2025-06-20
Check if serverTiming is defined before gathering its metrics.

gmierz:
This patch fixes an issue that occurs when taking a measurement with no navigations. In those cases, the `serverTiming` is undefined. Example for reproducing the failure:
```
module.exports = async function (context, commands) {
  context.log.info("Starting constant value regression test");
  await commands.measure.start("regression-test");
  await commands.measure.stop();
  await commands.measure.addObject({
    data: { metric: 1000 },
  });
};
```

### DIFF
--- a/browserscripts/timings/serverTimings.js
+++ b/browserscripts/timings/serverTimings.js
@@ -1,12 +1,14 @@
 (function () {
     let t = window.performance.getEntriesByType('navigation')[0];
     const serverTimings = [];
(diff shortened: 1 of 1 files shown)


soulgalore:
Thank you @gmierz !

gmierz:
Thanks for the quick review, and merge @soulgalore :)

PR #2320, opened 2025-09-16
Use new Firefox PerfStats API

dpalmeiro:
The existing perf stats API accepts a 32-bit bitmask so it only supports up to 32 features.  We've recently exceeded that limit and need to update this API.  In https://bugzilla.mozilla.org/show_bug.cgi?id=1981299 I've changed the API to take an array of feature strings instead of a bitmask.  

Browsertime will need to update to use the new API, but I've left backwards compatibility to the older interface as well.

### DIFF
--- a/lib/firefox/perfStats.js
+++ b/lib/firefox/perfStats.js
@@ -6,15 +6,51 @@ export class PerfStats {
     this.runner = runner;
   }
 
-  async start(featureMask) {
+  async start(features) {
     const runner = this.runner;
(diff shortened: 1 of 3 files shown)


soulgalore:
Cool, is this ready to go @dpalmeiro ?

dpalmeiro:
Yup, we'll need this change first before we can land bug [1981299](https://bugzilla.mozilla.org/show_bug.cgi?id=1981299).

soulgalore:
Cool, thank you @dpalmeiro !

